PR #26, opened 2021-08-29
flex was replaced with grid

io-med:
- [DEMO LINK](https://io-med.github.io/layout_catalog_grid/)


perronogrande (approved):

herkoss (changes requested):
Well done!👏👍
But there are too large margins
Gap between cards should be 48px

Fix it please

Screenshot 2021-08-31 at 13 32 11

herkoss (approved):
Great!👍😊

PR #69, opened 2021-09-19
Catalog was reworked

Igor-MateAcademy:
- [DEMO LINK](https://Igor-MateAcademy.github.io/layout_catalog_grid/)

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -1,9 +1,17 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
-
+  display: grid;
+  grid-template-columns: repeat(4, 1fr);
+  gap: 48px;
   max-width: 944px;
   margin: 0 auto;
+
+  @media (max-width: 1024px) {
+    grid-template-columns: repeat(3, 1fr);
+  }
+
+  @media (max-width: 768px) {
+    grid-template-columns: repeat(2, 1fr);
+  }
 
   &__card {
     display: flex;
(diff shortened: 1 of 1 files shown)


src/styles/_catalog.scss
```diff
@@ -1,10 +1,18 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
-
+  display: grid;
+  grid-template-columns: repeat(4, 1fr);
+  gap: 48px;
   max-width: 944px;
   margin: 0 auto;
```
podlesnyi-pavel, line 7: in the assignment we don't have to center element but i think it should be done

src/styles/_catalog.scss
```diff
@@ -1,10 +1,18 @@
...
 
+  @media (max-width: 1024px) {
+    grid-template-columns: repeat(3, 1fr);
+  }
+
+  @media (max-width: 768px) {
+    grid-template-columns: repeat(2, 1fr);
+  }
```
podlesnyi-pavel, line 14: There should be two columns starting at 488px

podlesnyi-pavel (changes requested):
- Ячейки грида расползаются и становятся больше карточек, образуя большие расстояния между карточками
- There should be two columns starting at 488px - условие не соблюдено 
- Думаю стоит центрировать карты внутри контейнера

podlesnyi-pavel (approved):
Good job
правда грид сетка еще расползается

VinnieJ-2k20 (changes requested):
1. Cards should be centered
image
2. On 3 or less cards in a row gap size is changing
image
image

VinnieJ-2k20 (changes requested):
In case of 3 or 2 cards in a row - gap size is changing. It should not.

Igor-MateAcademy:
> In case of 3 or 2 cards in a row - gap size is changing. It should not.

DEMO LINK fixed.

VinnieJ-2k20 (approved):

PR #90, opened 2021-09-21
done

AnastasiaDmitrivn:
- [DEMO LINK](https://AnastasiaDmitrivn.github.io/layout_catalog_grid/)

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -1,9 +1,21 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
+  // display: flex;
+  // flex-wrap: wrap;
+  display: grid;
+  justify-content: center;
+  grid-template-columns: repeat(auto-fill, minmax(200px, 1fr));
+  gap: 48px;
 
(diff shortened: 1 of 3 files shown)


src/styles/_catalog.scss
```diff
@@ -1,10 +1,22 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
+  // display: flex;
+  // flex-wrap: wrap;
```
misharosa, line 3: No need to leave comments

misharosa (changes requested):
Pick up comments, and finished=)

src/styles/_catalog.scss
AnastasiaDmitrivn: thanks)

VinnieJ-2k20 (changes requested):
In case of 3 or less cards in a row - gap size is changing. It should not.
image
image

Also, cards are not centered on the screen
image

VinnieJ-2k20 (approved):

PR #113, opened 2021-10-21
catalog grid

darynadudar:
- [DEMO LINK](https://DudarDaryna.github.io/layout_catalog_grid/)

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -4,10 +4,10 @@
   grid-template-columns: repeat(4, 200px);
   gap: 48px;
 
-  @media (max-width: 1024px) {
+  @media (max-width: 1023px) {
     grid-template-columns: repeat(3, 200px);
   }
-  @media (max-width: 768px) {
+  @media (max-width: 767px) {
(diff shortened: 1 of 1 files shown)


src/styles/_catalog.scss
```diff
@@ -1,22 +1,22 @@
...
+  gap: 48px;
 
-  max-width: 944px;
-  margin: 0 auto;
+  @media (max-width: 1024px) {
+    grid-template-columns: repeat(3, 200px);
+  }
+  @media (max-width: 768px) {
+    grid-template-columns: repeat(2, 200px);
+  }
```
Val-space, line 12: check your break points:) 4 cards appear from 1025 px ,not 1024px. same thing with 768px.
Screenshot 2021-10-21 at 21 27 23

Val-space (changes requested):
I think everything is good but needs a little adjustment with the media!

src/styles/_catalog.scss
DudarDaryna: hm, you are right!

PR #257, opened 2022-01-06
Develop

serhii-vorona:
- [DEMO LINK](https://serhii-vorona.github.io/layout_catalog_grid/)

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -1,8 +1,9 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
+  display: grid;
+  gap: 48px;
+  grid-template-columns: repeat(4, 1fr);
 
-  max-width: 944px;
+  max-width: 1024px;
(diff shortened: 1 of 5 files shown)


src/styles/_catalog.scss
```diff
@@ -1,8 +1,9 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
+  display: grid;
+  gap: 48px;
+  grid-template-columns: repeat(4, 1fr);
```
Anna-Tkachenko, line 4: You can also try to set auto-fill to grid-template-columns instead of using media-queries

src/styles/_catalog.scss
```diff
@@ -13,10 +14,18 @@
     height: 300px;
```
Anna-Tkachenko, line 14: It' better to use variables for main properties like card size

Anna-Tkachenko (approved):
Good job! 👍

PR #258, opened 2022-01-06
solution

013g:
https://013g.github.io/layout_catalog_grid/


src/styles/blocks/_catalog.scss
```diff
@@ -0,0 +1,29 @@
...
+    height: 300px;
+    border: 1px solid #ccc;
+    border-radius: 5px;
+    margin-bottom: 48px;
+  }
+
+}
+@media (max-width:768px) {
+  .catalog {
+    display: grid;
```
Anna-Tkachenko, line 20: Set display: grid only once for catalog, don't repeat it in each media query

Anna-Tkachenko (approved):

PR #266, opened 2022-01-10
solution

hserpukhovitina:


danylo-onishchuk (changes requested):
card's width is 200px including borders

Screenshot 2022-01-11 at 16 16 26

danylo-onishchuk (approved):

PR #284, opened 2022-01-27
flex was replased on a grid

antonganetskiy:
- [DEMO LINK](https://antonganetskiy.github.io/layout_catalog_grid/)

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -1,22 +1,27 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
+  display: grid;
+  justify-content: center;
+  gap: 48px;
 
-  max-width: 944px;
-  margin: 0 auto;
+  @media (min-width: 488px) {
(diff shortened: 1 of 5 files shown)


src/styles/_catalog.scss
```diff
@@ -1,22 +1,27 @@
...
+  @media (min-width: 1024px) {
+    grid-template-columns: repeat(4, 200px);
+  }
 
   &__card {
-    display: flex;
+    display: grid;
     justify-content: center;
     align-items: center;
     width: 200px;
```
jstmpelowycz, line 22: Code-style: I would suggest moving `width` value to a variable since it's been used several times

jstmpelowycz (approved):
Perfect, everything as it should be!

src/styles/_catalog.scss
antonganetskiy: ++ thenk you)

Ihor-Karpyn (approved):

PR #331, opened 2022-02-28
catalog-grid was created

maxxnikiforov:
- [DEMO LINK](https://maxxnikiforov.github.io/layout_catalog_grid/)

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -7,27 +7,21 @@
     display: grid;
     justify-content: center;
     align-items: center;
-    width: 200px;
+    width: $card-width;
     height: 300px;
     border: 1px solid #ccc;
     border-radius: 5px;
   }
-}
 
-@media (min-width: 488px) {
-  .catalog {
-    grid-template-columns: repeat(2, 200px);
+  @media (min-width: 488px) {
+    grid-template-columns: repeat(2, $card-width);
   }
-}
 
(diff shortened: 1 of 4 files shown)


src/styles/_catalog.scss
```diff
@@ -1,22 +1,33 @@
...
+    grid-template-columns: repeat(3, 200px);
+  }
+}
 
-  &__card:not(:nth-child(4n)) {
-    margin-right: 48px;
+@media (min-width: 1024px) {
+  .catalog {
+    grid-template-columns: repeat(4, 200px);
   }
```
Olexandr6, line 32: Use SCSS nesting, move these media queries inside their selector.

src/styles/_catalog.scss
```diff
@@ -1,22 +1,33 @@
...
+  display: grid;
+  gap: 48px;
+  justify-content: center;
 
   &__card {
-    display: flex;
+    display: grid;
     justify-content: center;
     align-items: center;
     width: 200px;
```
Olexandr6, line 10: Create a variable for card size value

Olexandr6: You use it in several places, so it's make sense to create a separate value

Olexandr6 (changes requested):
Almost done, but some small things should be fixed, check comments👇

illiarerun (approved):
Great job 🔥

PR #332, opened 2022-03-01
master, develop were merged

HelenPetrovska:
https://helenpetrovska.github.io/layout_catalog_grid/

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -1,22 +1,22 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
-
-  max-width: 944px;
-  margin: 0 auto;
+  display: grid;
+  grid-template-columns: repeat(auto-fill, 200px);
+  gap: 48px;
+  justify-content: center;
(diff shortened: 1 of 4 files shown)


Olexandr6 (changes requested):
Below 488px screen width, there should be only `1` card in the row:
image

In the range from 488px to 768px screen width, there should be only `2` cards in the row:
image

In the range from  768px  to 1024px screen width, there should be only `3` cards in the row:
image

src/styles/_catalog.scss
```diff
@@ -1,22 +1,29 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
-
-  max-width: 944px;
-  margin: 0 auto;
+  display: grid;
+  grid-template-columns: 200px;
```
Olexandr6, line 3: Create a separate variable for card width, cause you use it in several places, so it makes sense.

Olexandr6 (approved):
Good job🔥, one small thing should be optimized👇

PR #341, opened 2022-03-09
Develop

SeKaM0:
https://SeKaM0.github.io/layout_catalog_grid/

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -1,6 +1,6 @@
 .catalog {
   display: grid;
-  grid-template-columns: repeat(auto-fill, minmax(200px, 1fr));
+  justify-content: center;
   grid-gap: 48px;
   max-width: 944px;
   margin: 0 auto;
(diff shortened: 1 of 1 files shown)


src/styles/_catalog.scss
```diff
@@ -1,22 +1,19 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
-
+  display: grid;
+  grid-template-columns: repeat(auto-fill, minmax(200px, 1fr));
```
illiarerun, line 3: check task once again, your doesn't work properly:

src/styles/_catalog.scss
```diff
@@ -1,22 +1,19 @@
...
   max-width: 944px;
   margin: 0 auto;
 
   &__card {
     display: flex;
     justify-content: center;
     align-items: center;
+    align-self: center;
+    justify-self: center;
     width: 200px;
```
illiarerun, line 14: use variable for that value, you will reuse it

illiarerun (changes requested):
Nice start, let's work on fixes now:

illiarerun (approved):

PR #357, opened 2022-03-16
catalog-grid

DimaKotovich:
[DEMO](https://DimaKotovich.github.io/layout_catalog_grid/)

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -1,22 +1,34 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
-
-  max-width: 944px;
-  margin: 0 auto;
+  display: grid;
+  grid-template-columns: repeat(auto-fill,200px);
+  gap: 48px;
+  justify-content: center;
 
   &__card {
-    display: flex;
+    display: grid;
     justify-content: center;
     align-items: center;
     width: 200px;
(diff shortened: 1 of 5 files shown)


src/styles/_catalog.scss
```diff
@@ -1,22 +1,34 @@
...
+  grid-template-columns: repeat(auto-fill,200px);
+  gap: 48px;
+  justify-content: center;
 
   &__card {
-    display: flex;
+    display: grid;
     justify-content: center;
     align-items: center;
     width: 200px;
```
IlliaBezditnyi, line 11:
```suggestion
    width: 200px;
```
you should create variable for this value, cause you use it a lot in media queries

IlliaBezditnyi (approved):
Excellent job! 👍 You can upgrade it a bit :)

AlinaMatuschak (changes requested):
- from 768px to 1024px should be three columns
Screenshot 2022-03-17 at 09 39 03
- from 488px to 768px should be two columns
Screenshot 2022-03-17 at 09 40 03
- to 488px should be one column
Screenshot 2022-03-17 at 09 41 14

AlinaMatuschak (approved):
Something is broken on your demo link
But code looks fine 👌

PR #382, opened 2022-03-30
rekha

rekhakilari:
- [DEMO LINK](https://rekhakilari.github.io/layout_catalog_grid/)

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -20,3 +20,10 @@
     margin-right: 48px;
   }
 }
+$screen-sizes: "2" 488px, "3" 768px, "4" 1024px;
+
+@each $cards, $width in $screen-sizes {
+  @media (min-width: #{$width}) {
+    grid-template-columns: repeat(#{$cards}, 200px);
+  }
+}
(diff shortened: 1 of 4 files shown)


src/styles/_catalog.scss
```diff
@@ -20,3 +20,10 @@
     margin-right: 48px;
   }
 }
+$screen-sizes: "2" 488px, "3" 768px, "4" 1024px;
```
meforoma, line 23: list each option on a new line for better readability
```suggestion
$screen-sizes:
  "2" 488px,
  "3" 768px,
  "4" 1024px;
```

src/styles/_catalog.scss
```diff
@@ -20,3 +20,10 @@
...
+
```
meforoma, line 24: @rekhakilari you seem to have missed the main task
> Replace flex with grid

you still have flex

meforoma (changes requested):

src/styles/_catalog.scss
meforoma: and cards have margins - that should be replaced with gaps

PR #385, opened 2022-03-30
initial commit

amitpatiljc:
- [DEMO LINK](https://amitpatiljc.github.io/layout_catalog_grid/)

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -1,14 +1,25 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
+  display: grid;
+  justify-content: center;
+  grid-template-columns: repeat(1, 200px);
+  // margin: 0 auto;
+  gap: 48px;
 
-  max-width: 944px;
-  margin: 0 auto;
+  @media (min-width: 488px) {
+    grid-template-columns: repeat(2, 200px);
+  }
+
+  @media (min-width: 768px) {
+    grid-template-columns: repeat(3, 200px);
+  }
+
+  @media (min-width: 1024px) {
+    grid-template-columns: repeat(4, 200px);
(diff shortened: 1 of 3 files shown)


src/styles/_catalog.scss
```diff
@@ -1,22 +1,30 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
+  display: grid;
+  justify-content: center;
+  grid-template-columns: repeat(1, 200px);
+  // margin: 0 auto;
```
AndreiSuc, line 5: I've read somewhere in the suggestions that we should delete commented code.

AndreiSuc (approved):
Clean and understandable code. 👍

src/styles/_catalog.scss
meforoma: yep, better keep commits clean

src/styles/_catalog.scss
```diff
@@ -1,22 +1,30 @@
...
+    grid-template-columns: repeat(2, 200px);
+  }
+
+  @media (min-width: 768px) {
+    grid-template-columns: repeat(3, 200px);
+  }
+
+  @media (min-width: 1024px) {
+    grid-template-columns: repeat(4, 200px);
+  }
```
meforoma, line 18: @amitpatiljc for sake of practice you could try to generate all three `@media` using `@each` loop.
However it's not required.

meforoma (changes requested):
there are just two minor comments - and one is optional
otherwise you did great 👍

src/styles/_catalog.scss
amitpatiljc: Done

meforoma (approved):

PR #388, opened 2022-03-31
Grid catalog layout

iLebo-stack:
https://iLebo-stack.github.io/layout_catalog_grid/

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -1,22 +1,103 @@
-.catalog {
-  display: flex;
-  flex-wrap: wrap;
+$margin-reset: 0;
 
+.catalog {
+  display: grid;
+  grid-template-columns: repeat(4, 1fr);
+  gap: 48px;
+  justify-items: center;
   max-width: 944px;
   margin: 0 auto;
+  padding-bottom: 50px;
 
-  &__card {
-    display: flex;
-    justify-content: center;
-    align-items: center;
-    width: 200px;
-    height: 300px;
-    border: 1px solid #ccc;
-    border-radius: 5px;
-    margin-bottom: 48px;
+  @media (max-width: 1023px) {
+    grid-template-columns: repeat(3, 1fr);
   }
 
-  &__card:not(:nth-child(4n)) {
-    margin-right: 48px;
+  @media (max-width: 767px) {
+    grid-template-columns: repeat(2, 1fr);
+  }
+
+  &__card {
+    display: grid;
+    width: 200px;
+    border: 1px solid #ccc;
+    border-radius: 5px;
+    padding: 32px 16px 16px;
+
+    &-image {
+      width: 160px;
+      height: 134px;
+      margin: 0 3px 40px;
+    }
+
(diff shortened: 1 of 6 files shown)


src/styles/_catalog.scss
```diff
@@ -1,22 +1,103 @@
-.catalog {
-  display: flex;
-  flex-wrap: wrap;
+$margin-reset: 0;
```
andreisucman, line 1: I think that this variable is unnecessary. If you put margin as 0 it's automatically understood that they are reset. Naming a zero - reset is unnecessary work IMHO.

src/styles/_catalog.scss
```diff
@@ -1,22 +1,103 @@
...
   margin: 0 auto;
+  padding-bottom: 50px;
+
+  @media (max-width: 1023px) {
+    grid-template-columns: repeat(3, 1fr);
+  }
+
+  @media (max-width: 767px) {
+    grid-template-columns: repeat(2, 1fr);
+  }
```
andreisucman, line 18: It is not specified explicitly in the task, but I think you should have a one-row layout for mobile devices as well. At the moment your page has a 2-row layout for mobile devices, and even though the cards fit, they don't have any padding around them which makes the page look unbalanced. 

I suggest adding one more media query that will make cards stand in one row on mobile.

image

andreisucman (approved):
Hi Lebo. Your code satisfies the formal requirements but please have a look at my suggestions for a higher probability of fast approval.

src/styles/_catalog.scss
iLebo-stack: I could do that but the reason I did not do it in the first place is because the header breaks at mobile-screen width, and we aren't suppose to touch it

src/styles/_catalog.scss
```diff
@@ -1,22 +1,105 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
-
+  display: grid;
+  grid-template-columns: repeat(4, 1fr);
+  gap: 48px;
+  justify-items: center;
   max-width: 944px;
   margin: 0 auto;
```
ihor-jpeg, line 7: no external styles on blocks

src/styles/_catalog.scss
ihor-jpeg: this is not a responsive design

ihor-jpeg: i mean this one, does not have to be resposive

src/styles/_catalog.scss
```diff
@@ -1,22 +1,105 @@
...
+    &-image {
+      width: 160px;
+      height: 134px;
+      margin: 0 3px 40px;
+    }
+
+    &-title {
+      font-size: 12px;
+      font-weight: 500;
+      color: #060b35;
```
ihor-jpeg, line 38: variable?

ihor-jpeg (changes requested):
Good job! Take a look at the comments

ihor-jpeg (approved):
Good job!

PR #391, opened 2022-03-31
catalog grid

Sardorbek1122:
- [DEMO LINK](https://Sardorbek1122.github.io/layout_catalog_grid/)

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -23,6 +23,5 @@
     height: 300px;
     border: 1px solid #ccc;
     border-radius: 5px;
-    margin-bottom: 48px;
   }
 }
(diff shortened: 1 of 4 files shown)


src/styles/_catalog.scss
```diff
@@ -1,10 +1,20 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
+  display: grid;
+  justify-content: center;
+  grid-template-columns: repeat(1, 200px);
+  gap: 48px;
```
meforoma, line 5: it's generally ok, yet better to replace margins with gap
currently you've added gaps on top of margins
Image 2022-04-01 14 24 48

meforoma (changes requested):

src/styles/_catalog.scss
```diff
@@ -1,10 +1,20 @@
...
+  @media (min-width: 488px) {
+    grid-template-columns: repeat(2, 200px);
+  }
+
+  @media (min-width: 768px) {
+    grid-template-columns: repeat(3, 200px);
+  }
 
+  @media (min-width: 1024px) {
+    grid-template-columns: repeat(4, 200px);
```
ihor-jpeg, line 16: maybe a variable for card width? maybe use repeat()?

ihor-jpeg (approved):
Good job! Take a look at my comments

PR #401, opened 2022-04-03
Catalog grid task

proznerka:
- [DEMO LINK](https://proznerka.github.io/layout_catalog_grid/)

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -1,22 +1,24 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
+  display: grid;
+  justify-content: center;
+  gap: 48px;
+  grid-template-columns: repeat(2, 200px);
 
-  max-width: 944px;
-  margin: 0 auto;
+  @media (min-width: 768px) {
+    grid-template-columns: repeat(3, 200px);
+  }
+
+  @media (min-width: 1024px) {
+    grid-template-columns: repeat(4, 200px);
+  }
 
   &__card {
-    display: flex;
+    display: grid;
     justify-content: center;
     align-items: center;
     width: 200px;
     height: 300px;
     border: 1px solid #ccc;
     border-radius: 5px;
-    margin-bottom: 48px;
-  }
-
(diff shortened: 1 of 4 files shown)


src/styles/_catalog.scss
```diff
@@ -1,22 +1,24 @@
...
+    grid-template-columns: repeat(3, 200px);
+  }
+
+  @media (min-width: 1024px) {
+    grid-template-columns: repeat(4, 200px);
+  }
 
   &__card {
-    display: flex;
+    display: grid;
```
andreisucman, line 16: I think you need to remove `display: grid` from the card, because it makes the card the grid of its own, which is not necessary here. The card is an element of the `.catalog` grid, but not a grid of its own.

src/styles/_catalog.scss
```diff
@@ -1,22 +1,24 @@
...
 
-  max-width: 944px;
-  margin: 0 auto;
+  @media (min-width: 768px) {
+    grid-template-columns: repeat(3, 200px);
+  }
+
+  @media (min-width: 1024px) {
+    grid-template-columns: repeat(4, 200px);
+  }
```
andreisucman, line 13: It might not seem obvious, but in the requirements, it says `There should be two columns starting at 488px `so I guess that below 488px there should be 1 column layout.

andreisucman (changes requested):
Hi Karina. Please consider implementing these changes for faster approval.

andreisucman (approved):
Good job.

src/styles/_catalog.scss
```diff
@@ -1,9 +1,20 @@
...
+    grid-template-columns: repeat(2, 200px);
+  }
+
+  @media (min-width: 768px) {
+    grid-template-columns: repeat(3, 200px);
+  }
+
+  @media (min-width: 1024px) {
+    grid-template-columns: repeat(4, 200px);
+  }
```
ihor-jpeg, line 17: maybe use repeat() and just one media query?

ihor-jpeg (approved):
Good job! Take a look at me comment

PR #406, opened 2022-04-06
catalog_grid

dima-pyvovarchuk:
[DEMO LINK](https://dima-pyvovarchuk.github.io/layout_catalog_grid/)


src/styles/main.scss
```diff
@@ -12,3 +12,22 @@ body {
...
+  }
+
+  @media (min-width: 768px) {
+    grid-template-columns: repeat(3, 200px);
+  }
+
+  @media (min-width: 1024px) {
+    grid-template-columns: repeat(4, 200px);
+  }
+}
```
Bohdan-Maksymiuk, line 33: All good, but why did you write it here. You already have `src/styles/_catalog.scss`. You should write it here

Bohdan-Maksymiuk (changes requested):

src/styles/main.scss
dima-pyvovarchuk: Maybe I misunderstood this rule)

But I fixed files

PR #415, opened 2022-04-11
task was done

alexandra-protyanova:


src/styles/_catalog.scss
```diff
@@ -1,22 +1,33 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
+  display: grid;
+  grid-template-columns: repeat(1, 200px);
```
vlad-demchuk, line 3: do you need repeat for one column?

vlad-demchuk (approved):
use descriptive names for commits

Denchepornyuk (approved):

PR #488, opened 2022-05-19
catalog with grid

Prince3378:
- [DEMO LINK](https://Prince3378.github.io/layout_catalog_grid/)

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -1,3 +1,5 @@
+$widthCol: 488px  2, 768px  3, 1024px  4;
+
 @mixin forGrid($column) {
   grid-template-columns: repeat($column, 200px);
   display: grid;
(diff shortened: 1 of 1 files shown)


NikitaKarbazin:
Great job, but I would do it through `@each` , instead of writing breakpoints, I would make a map with two values, one with the width of the screen, the other with the number of columns (there will be as many as we have breakpoints), then we can optimize our code, make it more elegant! Look at the documentation and maybe you can fing good variation for your project! But all was work and i approve!
[SASS each]( https://sass-lang.com/documentation/at-rules/control/each)

NikitaKarbazin (approved):

src/styles/_catalog.scss
```diff
@@ -1,22 +1,30 @@
+$widthCol: 488px  2, 768px  3, 1024px  4;
```
illiarerun, line 1: format it properly, put each pair on a newline

src/styles/_catalog.scss
```diff
@@ -1,22 +1,30 @@
...
+
+@mixin forGrid($column) {
```
illiarerun, line 3:
```suggestion
@mixin catalogGrid($column-count) {
```

illiarerun (approved):
Good job, but you need to fix these:

PR #556, opened 2022-06-10
Develop

vitalii-borzov:
- [DEMO LINK](https://vitalii-borzov.github.io/layout_catalog_grid/)


Bohdan-Maksymiuk (changes requested):
> There should be four columns starting at 1024px
Screenshot 2022-06-14 at 14 07 18

and check other media from task - they don't work as they should

aliceblunt8 (approved):

PR #592, opened 2022-06-18
created catalog page using grid

Sergii2705:
[DEMO LINK](https://Sergii2705.github.io/layout_catalog_grid/)


src/styles/blocks/catalog.scss
```diff
@@ -0,0 +1,22 @@
...
+    @media (min-width: $resolution) {
+      grid-template-columns: repeat($column, $card-width);
+    }
+  }
+
+  &__card {
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    width: 200px;
```
solaryasha, line 17: You don't need to hardcode card width. Remove this rule

solaryasha (changes requested):

src/styles/blocks/catalog.scss
Sergii2705: OK got it

VETALEK (changes requested):
Code is good, need to deploy:

image

PR #599, opened 2022-06-19
flex was changed to grid

tetyanabukoros:
- [DEMO LINK](https://tetyanabukoros.github.io/layout_catalog_grid/)


src/styles/_catalog.scss
```diff
@@ -1,6 +1,20 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
+  display: grid;
+  justify-items: center;
+  grid-template-columns: repeat(4, minmax(200px, 1fr));
+  gap: 48px;
+
+  @media (max-width: 1024px) {
+    grid-template-columns: repeat(3, minmax(200px, 1fr));
```
yuliia-dronova, line 8: I'm not sure that you have to use **minmax** here wth 1fr.. that makes big gaps between cards with changing screen's size:

image

It is  better to use fixed cell's width: 200px

src/styles/_catalog.scss
```diff
@@ -1,6 +1,20 @@
...
+
+  @media (max-width: 768px) {
+    grid-template-columns: repeat(2, minmax(200px, 1fr));
+  }
+
+  @media (max-width: 488px) {
+    grid-template-columns: repeat(1, minmax(200px, 1fr));
+  }
 
   max-width: 944px;
```
yuliia-dronova, line 19: When you fix rules with minmax, you can just delete max-width for block ;)

src/styles/_catalog.scss
```diff
@@ -1,6 +1,20 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
+  display: grid;
+  justify-items: center;
+  grid-template-columns: repeat(4, minmax(200px, 1fr));
+  gap: 48px;
+
+  @media (max-width: 1024px) {
```
yuliia-dronova, line 7: Optional: you can use cycle for media just for convenience)

yuliia-dronova (changes requested):

yuliia-dronova (approved):

src/styles/_catalog.scss
tetyanabukoros: Unfortunately, I could not use cycle, it doesn't work for some reason

VETALEK (approved):
Fire 🔥

src/styles/_catalog.scss
VETALEK: looks at others' cycles, if you made the same and it still doesn't work - good reason to visit QnA 😉

PR #642, opened 2022-07-04
Initial commit

Vasyliy22:
- [DEMO LINK](https://Vasyliy22.github.io/layout_catalog_grid/)


ArturErmolenko (changes requested):

src/styles/_catalog.scss
```diff
@@ -1,22 +1,48 @@
+// * {
+//   box-sizing: border-box;
+// }
```
Bohdan-Maksymiuk, line 3: remove comments

src/styles/_catalog.scss
```diff
@@ -1,22 +1,48 @@
...
 }
+
+@media (max-width: 488px) {
+  .catalog {
+    justify-content: center;
+    grid-template-columns: repeat(1, 200px);
+  }
+}
+
+@media (min-width: 488px) {
```
Bohdan-Maksymiuk, line 36: 😕

Bohdan-Maksymiuk (changes requested):
> There should be four columns starting at 1024px
Screenshot 2022-07-12 at 11 23 40

You should use `one-direction` approach for media-queries
For example:
* write default (it will be used before 488px)
* then media `min-width: 488px`
* then media `min-width: 768px`
* then media `min-width: 1024px`

alexandra-protyanova (approved):

PR #662, opened 2022-07-13
catalog_grid

viktoria-krasnyuk:
[DEMO LINK](https://viktoria-krasnyuk.github.io/layout_catalog_grid/)


Bohdan-Maksymiuk (changes requested):
> Gap between cards should be 48px
Screenshot 2022-07-16 at 13 45 46
I guess there is more space between cards here

PR #771, opened 2022-08-06
commit message

AMahalias:
[DEMO LINK](https://AMahalias.github.io/layout_catalog_grid/)

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -1,9 +1,10 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
-
-  max-width: 944px;
+  display: grid;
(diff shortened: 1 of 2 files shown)


src/styles/_catalog.scss
```diff
@@ -13,10 +14,23 @@
     height: 300px;
     border: 1px solid #ccc;
     border-radius: 5px;
-    margin-bottom: 48px;
   }
+}
+
+@media (min-width: 488px) and (max-width: 767px) {
```
alexandra-protyanova, line 20: it is better to use or max or min width. In this task there is no need in combination of rules for media

alexandra-protyanova (changes requested):

alexandra-protyanova (approved):

PR #804, opened 2022-08-19
catalog_grid

MaxManis:
- [DEMO LINK](https://MaxManis.github.io/layout_catalog_grid/)

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -2,19 +2,18 @@ $card-height: 300px;
 $card-width: 200px;
 $gap-size: 48px;
 $catalog-max-width: $card-width * 4 + $gap-size * 3;
+$text-hover-color: #34568b;
 
 .catalog {
   display: grid;
   justify-content: center;
+  align-content: center;
   justify-items: center;
   align-items: center;
 
-  grid-template-columns: 1fr 1fr 1fr 1fr;
+  grid-template-columns: repeat(4, $card-width);
   grid-auto-rows: $card-height;
   gap: $gap-size;
-
-  max-width: $catalog-max-width;
-  min-width: $card-width;
 
   &__card {
     display: flex;
@@ -24,6 +23,14 @@ $catalog-max-width: $card-width * 4 + $gap-size * 3;
     height: $card-height;
     border: 1px solid #ccc;
     border-radius: 5px;
+
+    transform-origin: center;
+    transition: color 0.3s, transform 0.3s;
+
+    &:hover {
+      color: $text-hover-color;
+      transform: scale(1.2);
+    }
   }
 
   @media (max-width: 487px) {
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -1,22 +1,38 @@
-.catalog {
-  display: flex;
-  flex-wrap: wrap;
+$card-height: 300px;
+$card-width: 200px;
+$gap-size: 48px;
+$catalog-max-width: $card-width * 4 + $gap-size * 3;
 
-  max-width: 944px;
-  margin: 0 auto;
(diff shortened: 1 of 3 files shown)


src/styles/_catalog.scss
```diff
@@ -1,22 +1,45 @@
...
+$text-hover-color: #34568b;
+
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
+  display: grid;
+  justify-content: center;
+  align-content: center;
+  justify-items: center;
+  align-items: center;
```
MaksimusCRS, line 12: Do you need all this rules?

src/styles/_catalog.scss
```diff
@@ -1,22 +1,45 @@
...
-  &__card:not(:nth-child(4n)) {
-    margin-right: 48px;
+  @media (max-width: 487px) {
+    grid-template-columns: 1fr;
+  }
+  @media (min-width: 488px) and (max-width: 767px) {
+    grid-template-columns: 1fr 1fr;
+  }
+  @media (min-width: 768px) and (max-width: 1023px) {
+    grid-template-columns: 1fr 1fr 1fr;
```
MaksimusCRS, line 43: You can use only 1 media query with one constraint

MaksimusCRS: Hint: try to use auto-fill

MaksimusCRS (changes requested):

src/styles/_catalog.scss
MaxManis: @MaksimusCRS yeah,  ` justify-content: center;` puts my grid in the center of the screan if it wider than the grid width, but not put it on the left side)
and `justify-items: center;` makes the same for each card, if screan includes 3 card only you will see it on the left side but  not on the center)

MaksimusCRS: @MaxManis OK, and what about align-content, why do you need it?)

MaxManis: @MaksimusCRS yeah, agreed. `align-*` rules doesnt really helpful here because of the same sizes for row and content)

src/styles/_catalog.scss
MaxManis: @MaksimusCRS hmm, interesting)) i could try XD)

MaxManis: @MaksimusCRS so did you mean this:
```
  @media (max-width: 1023px) {
    grid-template-columns: repeat(auto-fill, $card-width);
  }
```
?))

MaxManis: image
in this case i have only one `@media` query but i have other problems, as on the picture)

MaxManis: other comments i fixed =)

MaksimusCRS: Ok, so you can add one more media query and it would be better then your previous solution :slightly_smiling_face:

MaksimusCRS: Actually you don't need 2 constraints anywhere, it's main mistake here

MaxManis: @MaksimusCRS as i reallized and @OleksiiBlezniuk said on QnA - using min-max we creating a range of spesified breakpoints for different kinds of divice, same if i dont use the exact min-max for other diveces it wont work correctly, like it will change my grid because of content size but not of screen size) So I choosed this way just because its sounds right and the result will be more expected, am i wrong?)

MaxManis: @MaksimusCRS yeah, I agree i could use `@media (max-width: 767px)` without `and  (min-width: 488px)`  and it will work absolutley the same)

MaksimusCRS (approved):
GJ

PR #806, opened 2022-08-20
grid layout implemented

jandrus-lviv:
[DEMO LINK](https://jandrus-lviv.github.io/layout_catalog_grid/)


liliia-livandovska (changes requested):
Good job! Please add center alignment of cards, so it should look like this

Знімок екрана 2022-08-23 о 11 45 54

src/styles/blocks/_catalog.scss
```diff
@@ -0,0 +1,23 @@
...
+    grid-template-columns: repeat(2, #{$card-width});
+  }
+
+  @media (min-width: 768px) {
+    grid-template-columns: repeat(3, #{$card-width});
+  }
+
+  @media (min-width: 1024px) {
+    grid-template-columns: repeat(4, #{$card-width});
+  }
```
liliia-livandovska, line 16: great job, you can also try to use mixin there like this if you want  @mixin col-width($width, $col) {
  @media (min-width: $width) {
    grid-template-columns: repeat($col, $card-width);
  }
}

jandrus-lviv: Thanks! Now I better understand how mixins work :)

marina-tilniak (approved):

PR #810, opened 2022-08-21
First Commit

rstarovoit:


package.json
```diff
@@ -21,7 +21,7 @@
     "@mate-academy/bemlint": "^0.1.1",
```
grafit-off, line 21: Don't add this type of files to git next time.

src/styles/_catalog.scss
```diff
@@ -1,9 +1,10 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
-
-  max-width: 944px;
-  margin: 0 auto;
+  display: grid;
+  grid-auto-flow: row;
+  justify-content: center;
+  grid-template-columns: repeat(4, 200px);
```
grafit-off, line 5: You can use variables for card-width, in reason that this value was using several times (in catalog and in catalog__card)

grafit-off (approved):
Good job! Keep going! Don't forget to add links to pull request description!

Anna-Tkachenko (approved):
⭐️

PR #813, opened 2022-08-22
Develop

grafit-off:
[DEMO LINK](https://grafit-off.github.io/layout_catalog_grid/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -2,50 +2,109 @@
 <html lang="en">
   <head>
     <meta charset="UTF-8">
-    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <meta
+      name="viewport"
+      content="width=device-width, initial-scale=1.0"
+    >
     <title>Product</title>
 
-    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link
+      rel="preconnect"
+      href="https://fonts.googleapis.com"
+    >
     <link
       rel="preconnect"
       href="https://fonts.gstatic.com"
       crossorigin="use-credentials"
     >
-    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@300;400;500&display=swap" rel="stylesheet">
+    <link
+      href="https://fonts.googleapis.com/css2?family=Roboto:wght@300;400;500&display=swap"
+      rel="stylesheet"
+    >
 
-    <link rel="stylesheet" href="./styles/main.scss">
+    <link
+      rel="stylesheet"
+      href="./styles/main.scss"
+    >
   </head>
   <body>
     <header class="header">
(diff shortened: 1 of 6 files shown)


src/styles/_normalizer.scss
```diff
@@ -0,0 +1,11 @@
...
+  font-family: Roboto, sans-serif;
+}
+
+body {
+  margin: 0;
+}
+
+* {
+  box-sizing: border-box;
+}
```
rstarovoit, line 11: I guess you should better place these changes to the top as it applies to all elements

src/styles/_catalog.scss
```diff
@@ -1,6 +1,8 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
+  display: grid;
+  grid-template-columns: repeat(auto-fill, $card-width);
+  justify-content: center;
+  gap: 48px;
 
   max-width: 944px;
   margin: 0 auto;
```
rstarovoit, line 8: Do you think these styles still matter after applying grid layout?

src/index.html
```diff
@@ -2,50 +2,109 @@
...
+    <link
+      href="https://fonts.googleapis.com/css2?family=Roboto:wght@300;400;500&display=swap"
+      rel="stylesheet"
+    >
 
-    <link rel="stylesheet" href="./styles/main.scss">
+    <link
+      rel="stylesheet"
+      href="./styles/main.scss"
+    >
```
rstarovoit, line 28: I do not think you should change it every time when you have a code line length below 80, not long attribute values and the number of attributes is less than 2. This comment for all such cases.

From Check List:

[CODE STYLE] - Keep your code line length **_below 80_**. It’s not only historical tradition, but also allows your code to fit into one standard screen without horizontal scroll. But do not break the line if it cannot be broken (ex., long links).

[CODE STYLE] - If the HTML-element has long attribute values or number of attributes is **_more than 2_** - start each one, including the first, on the new line with 2-space indentation related to tag. Tag’s closing bracket should be on the same level as opening one.

rstarovoit (changes requested):
Nice work!

src/index.html
TheVovchik: it's good practice, because if he added here class atribute it's already styled good. So agree with initial code

src/styles/_catalog.scss
TheVovchik: Yes it center Grid container in the page.

rstarovoit: For my case "justify-content: center" is enoght.
https://youtu.be/66pn_kRNMto?t=92

MaksimusCRS (approved):
GJ

PR #815, opened 2022-08-22
flex was replaced with grid

TheVovchik:
- [DEMO LINK](https://TheVovchik.github.io/layout_catalog_grid/)

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -1,22 +1,33 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
+  display: grid;
+  gap: $cards-gap;
+  grid-template-columns: repeat(auto-fill, $card-width);
+  justify-content: center;
+  max-width: $four-cards-container;
 
-  max-width: 944px;
-  margin: 0 auto;
+  &--margins--bottom {
+    @include margins(auto);
+
+    @media (max-width: 1024px) {
+      @include margins(40px);
+    }
+
+    @media (max-width: 788px) {
+      @include margins(36px);
+    }
+
+    @media (max-width: 528px) {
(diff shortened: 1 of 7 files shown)


imaksweb (approved):
Good job!

silchyn (changes requested):
Slowly resize from one column to two, do you also see that jumping?

TheVovchik:
> Slowly resize from one column to two, do you also see that jumping?

what jump and when do you see? don't understand you. I have 1 column one column below 488px, two columns starting at 488px, three columns starting at 768px and 4 columns starting at 1024px

src/styles/_catalog.scss
```diff
@@ -1,22 +1,33 @@
...
+
+    @media (max-width: 1023.9px) {
+      grid-template-columns: repeat(3, $card-width);
+    }
+
+    @media (max-width: 767.9px) {
+      grid-template-columns: repeat(2, $card-width);
+    }
+
+    @media (max-width: 487.9px) {
```
silchyn, line 19: Where are these numbers from?

silchyn (changes requested):
Now there are no jumps, but

src/styles/_catalog.scss
TheVovchik: image
because of task. If I put 488px it will not change on two columns! 

image
same here!

image
same here!

TheVovchik: I saw one guy in Youtube he use such strange technique and it works. It's strange because I ever see screen 487.5px or smth like that, but in general for me it's more understandable that my media query is about 1024 when I use 1023.9 razer then I used to use 1023.

silchyn (approved):
Cool!

PR #819, opened 2022-08-23
grid-catalog

artem-musii:
- [DEMO LINK](https://artem-musii.github.io/layout_catalog_grid/)


Anna-Borodaienko (commented):
I'd recommend having different files for reset and normalize, and different folders for blocks and utils. Maybe it's not a mistake now, but we have taken this structure - and we must use it, I think.

And add informative links to HTML - like Anna showed today on the lesson.

Andr1yK (approved):

PR #824, opened 2022-08-24
catalog grid

ghost:
[DEMO LINK](https://kamal-tufekcic.github.io/layout_catalog_grid/)

Based on my understanding, we are only supposed to edit main.scss, so I deleted the rest of styles and put everything in there, the rest of the changes were implemented as instructed


src/styles/main.scss
```diff
@@ -10,5 +10,83 @@ body {
...
+    & {
+      grid-template-columns: repeat(2, 200px);
+    }
+  }
+
+  @media (max-width: 487px) {
+    & {
+      grid-template-columns: repeat(1, 200px);
+    }
+  }
```
VinnieJ-2k20, line 46: You don't need to use `&` for media in scss.
You can simply write under necessary selector.
Here you apply it to `.catalog`, so change these and put them right after line 18-19.

```
@media (condition) {
  some: styles;
}
```

VinnieJ-2k20 (changes requested):
Check the comment about scss and media.
Fix this and that's it.

VinnieJ-2k20 (approved):
Good job 👍

PR #830, opened 2022-08-25
Grid-layout for catalog was created

Yevheniya-Deryka:
- [DEMO LINK](https://Yevheniya-Deryka.github.io/layout_catalog_grid/)

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -1,22 +1,35 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
-
-  max-width: 944px;
-  margin: 0 auto;
+  display: grid;
+  grid-template-columns: repeat(1, 200px);
+  gap: 48px;
+  justify-content: center;
 
   &__card {
     display: flex;
     justify-content: center;
     align-items: center;
-    width: 200px;
+
     height: 300px;
+
     border: 1px solid #ccc;
     border-radius: 5px;
-    margin-bottom: 48px;
   }
+}
 
-  &__card:not(:nth-child(4n)) {
-    margin-right: 48px;
+@media (min-width: 488px) {
+  .catalog {
+    grid-template-columns: repeat(2, 200px);
+  }
+}
+
+@media (min-width: 768px) {
+  .catalog {
+    grid-template-columns: repeat(3, 200px);
+  }
+}
+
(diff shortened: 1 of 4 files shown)


src/styles/_catalog.scss
```diff
@@ -1,22 +1,35 @@
...
-
-  max-width: 944px;
-  margin: 0 auto;
+  display: grid;
+  grid-template-columns: repeat(1, 200px);
+  gap: 48px;
+  justify-content: center;
 
   &__card {
     display: flex;
```
volodymyrklymkiv, line 8: You can use here display: grid

src/styles/_catalog.scss
```diff
@@ -1,22 +1,35 @@
...
+
+@media (min-width: 768px) {
+  .catalog {
+    grid-template-columns: repeat(3, 200px);
+  }
+}
 
-  &__card:not(:nth-child(4n)) {
-    margin-right: 48px;
+@media (min-width: 1024px) {
```
volodymyrklymkiv, line 31: also u can use just one media query for 488px, if u will put "auto-fill" instead of number in "grid template" , but i think its ok )

src/styles/_catalog.scss
```diff
@@ -1,22 +1,35 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
-
-  max-width: 944px;
-  margin: 0 auto;
+  display: grid;
+  grid-template-columns: repeat(1, 200px);
```
volodymyrklymkiv, line 3: u can add variable for 200px

volodymyrklymkiv (approved):

src/styles/_catalog.scss
Yevheniya-Deryka: then I'll have more than 4 columns for wide screen

src/styles/_catalog.scss
Yevheniya-Deryka: yep, but thought, it's not significant in this task)

src/styles/_catalog.scss
Yevheniya-Deryka: yes, but now I have only one title, so no need to do that

MaksimusCRS (approved):
GJ

PR #832, opened 2022-08-25
added grid and variables

volodymyrklymkiv:
- [DEMO LINK](https://volodymyrklymkiv.github.io/layout_catalog_grid/)

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -1,22 +1,26 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
+  display: grid;
+  gap: 48px;
+  justify-content: center;
+  grid-template-columns: repeat(auto-fill, $width);
 
   max-width: 944px;
   margin: 0 auto;
 
   &__card {
-    display: flex;
-    justify-content: center;
+    display: grid;
+    justify-items: center;
     align-items: center;
-    width: 200px;
+    width: $width;
     height: 300px;
     border: 1px solid #ccc;
     border-radius: 5px;
-    margin-bottom: 48px;
+    margin-bottom: $gap;
   }
+}
 
-  &__card:not(:nth-child(4n)) {
-    margin-right: 48px;
+@media (max-width: 488px) {
+  .catalog {
+    grid-template-columns: repeat(auto-fit, 1fr);
   }
 }
--- a/readme.md
+++ b/readme.md
@@ -1,6 +1,6 @@
 # Frontend practice with catalog page using grid
 Replace `<your_account>` with your Github username and copy the links to Pull Request description:
-- [DEMO LINK](https://<your_account>.github.io/layout_catalog_grid/)
+- [DEMO LINK](https://volodymyrklymkiv.github.io/layout_catalog_grid/)
 
 > Follow [this instructions](https://github.com/mate-academy/layout_task-guideline#how-to-solve-the-layout-tasks-on-github)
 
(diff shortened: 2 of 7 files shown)


src/styles/_catalog.scss
```diff
@@ -1,22 +1,26 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
+  display: grid;
+  gap: 48px;
+  justify-content: center;
+  grid-template-columns: repeat(auto-fill, $width);
 
   max-width: 944px;
```
Yevheniya-Deryka, line 7: Don't think you need to use max-width here in this task

src/styles/_catalog.scss
```diff
@@ -1,22 +1,26 @@
...
   margin: 0 auto;
```
Yevheniya-Deryka, line 8: Don't sure we can use margins for BEM-block

src/styles/_catalog.scss
```diff
@@ -1,22 +1,26 @@
...
+    display: grid;
+    justify-items: center;
     align-items: center;
-    width: 200px;
+    width: $width;
     height: 300px;
     border: 1px solid #ccc;
     border-radius: 5px;
-    margin-bottom: 48px;
+    margin-bottom: $gap;
```
Yevheniya-Deryka, line 18: If you use gap for the container, you don't need this margin here, cause you get the wrong vertical-gap between cards

src/styles/_catalog.scss
```diff
@@ -1,22 +1,26 @@
...
+  justify-content: center;
+  grid-template-columns: repeat(auto-fill, $width);
 
   max-width: 944px;
   margin: 0 auto;
 
   &__card {
-    display: flex;
-    justify-content: center;
+    display: grid;
```
Yevheniya-Deryka, line 11: why do you use grid here?

src/styles/_catalog.scss
```diff
@@ -1,22 +1,26 @@
...
     border: 1px solid #ccc;
     border-radius: 5px;
-    margin-bottom: 48px;
+    margin-bottom: $gap;
   }
+}
 
-  &__card:not(:nth-child(4n)) {
-    margin-right: 48px;
+@media (max-width: 488px) {
```
Yevheniya-Deryka, line 22: You should have 1 column under 488px, but you have 2 columns at 450px. Check your media queries for different width

Yevheniya-Deryka (changes requested):
Left some comments

src/styles/_catalog.scss
Andr1yK: U don't need to hard-code :(

But, U can use variables and calculate the value with the formula

src/styles/_catalog.scss
```diff
@@ -1,22 +1,26 @@
...
     height: 300px;
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
   }
+}
 
-  &__card:not(:nth-child(4n)) {
-    margin-right: 48px;
+@media (max-width: 488px) {
```
Andr1yK, line 22: Media queries
Now: U show 4 columns from 960px
image

Now: U show 3 columns from 712px
image

Now: U show 2 columns from 464px
image

There should be one column below 488px
And about this point, u show one column below 464px

Andr1yK (changes requested):
Good start, but u really need to fix media queries 😄

Andr1yK (approved):

PR #836, opened 2022-08-26
added a grid for catalog

vladimir-dublya:
- [DEMO LINK](https://vladimir-dublya.github.io/layout_catalog_grid/)

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -1,22 +1,46 @@
-.catalog {
-  display: flex;
-  flex-wrap: wrap;
+$card-height: 300px;
+$card-width: 200px;
+$gap-size: 48px;
+$cards-counts: 4;
+$gap-counts: $cards-counts - 1;
+$catalog-max-width: $card-width * $cards-counts + $gap-size * $gap-counts;
+$text-hover-color: #34568b;
 
-  max-width: 944px;
-  margin: 0 auto;
+.catalog {
+  display: grid;
+  justify-content: center;
+  justify-items: center;
+
+  grid-template-columns: repeat(4, $card-width);
+  grid-auto-rows: $card-height;
+  gap: $gap-size;
 
   &__card {
+
     display: flex;
     justify-content: center;
     align-items: center;
-    width: 200px;
-    height: 300px;
+    width: $card-width;
+    height: $card-height;
     border: 1px solid #ccc;
     border-radius: 5px;
-    margin-bottom: 48px;
+
+    transform-origin: center;
+    transition: color 0.3s, transform 0.3s;
+
+    &:hover {
+      color: $text-hover-color;
+      transform: scale(1.2);
+    }
   }
 
-  &__card:not(:nth-child(4n)) {
-    margin-right: 48px;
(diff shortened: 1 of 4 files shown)


src/styles/_catalog.scss
```diff
@@ -1,22 +1,46 @@
...
     align-items: center;
-    width: 200px;
-    height: 300px;
+    width: $card-width;
+    height: $card-height;
     border: 1px solid #ccc;
     border-radius: 5px;
-    margin-bottom: 48px;
+
+    transform-origin: center;
```
imaksweb, line 28: I think you can remove this rule, cause it's default value

src/styles/_catalog.scss
```diff
@@ -1,22 +1,46 @@
+$card-height: 300px;
+$card-width: 200px;
+$gap-size: 48px;
+$cards-counts: 4;
+$gap-counts: $cards-counts - 1;
+$catalog-max-width: $card-width * $cards-counts + $gap-size * $gap-counts;
```
imaksweb, line 6: It seems like you don't use this variable at all

src/styles/_catalog.scss
```diff
@@ -1,22 +1,46 @@
...
-  &__card:not(:nth-child(4n)) {
-    margin-right: 48px;
+  @media (max-width: 1023px) {
+    grid-template-columns: repeat(3, $card-width);
+  }
+  @media (max-width: 767px) {
+    grid-template-columns: repeat(2, $card-width);
+  }
+  @media (max-width: 487px) {
+    grid-template-columns: $card-width;
```
imaksweb, line 44: I'm not sure that we should strictly follow breakpoints and create all this media queries instead of using just one(we can use auto-fill). But I'm convinced it's completely optional.

src/styles/_catalog.scss
```diff
@@ -1,22 +1,46 @@
...
-  display: flex;
-  flex-wrap: wrap;
+  display: grid;
+  justify-content: center;
+  justify-items: center;
 
-  max-width: 944px;
-  margin: 0 auto;
+  grid-template-columns: repeat(4, $card-width);
+  grid-auto-rows: $card-height;
```
imaksweb, line 15: As for me you can remove this rule, it does nothing

src/styles/_catalog.scss
```diff
@@ -1,22 +1,46 @@
...
+$gap-counts: $cards-counts - 1;
+$catalog-max-width: $card-width * $cards-counts + $gap-size * $gap-counts;
+$text-hover-color: #34568b;
+
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
+  display: grid;
+  justify-content: center;
+  justify-items: center;
```
imaksweb, line 12: Are you sure you need this line?

imaksweb (approved):
GJ!

src/styles/_catalog.scss
Denchepornyuk: yep, auto-fill should work here, with media queries also good solution

Denchepornyuk (approved):

PR #842, opened 2022-08-26
catalog_grid

vallerii:
- [DEMO LINK](https://vallerii.github.io/layout_catalog_grid/)


src/styles/_catalog.scss
```diff
@@ -1,12 +1,26 @@
...
+    grid-template-columns: repeat(2, 200px);
+  }
+
+  @media (min-width: 768px) {
+    grid-template-columns: repeat(3, 200px);
+  }
+
+  @media (min-width: 1024px) {
+    grid-template-columns: repeat(4, 200px);
+  }
```
imaksweb, line 20: You can use css variable for columns count and change only the value of this variable depending on resolution, but it's completely optional

imaksweb (changes requested):
It seems more than 48px gap
img1

src/styles/_catalog.scss
```diff
@@ -1,22 +1,35 @@
...
     height: 300px;
     border: 1px solid #ccc;
     border-radius: 5px;
-    margin-bottom: 48px;
   }
 
   &__card:not(:nth-child(4n)) {
-    margin-right: 48px;
+    gap: 48px;
   }
```
imaksweb, line 34: This rule does nothing, cause gap add indentations only between grid elements, so you can remove it at all

src/styles/_catalog.scss
```diff
@@ -1,22 +1,35 @@
...
+  @media (min-width: 1024px) {
+    grid-template-columns: repeat(4, 200px);
+  }
 
   &__card {
-    display: flex;
+    display: grid;
     justify-content: center;
     align-items: center;
     width: 200px;
```
imaksweb, line 26: You don't need this rule as well, you already set a width for cards in grid-template-columns

imaksweb (changes requested):

imaksweb (approved):
Good job 🔥

src/styles/_catalog.scss
```diff
@@ -1,22 +1,31 @@
...
+  @media (min-width: 488px) {
+    grid-template-columns: repeat(2, 200px);
+  }
+
+  @media (min-width: 768px) {
+    grid-template-columns: repeat(3, 200px);
+  }
+
+  @media (min-width: 1024px) {
+    grid-template-columns: repeat(4, 200px);
```
silchyn, line 19: What is 200?

silchyn (changes requested):
Good job, but

src/styles/_catalog.scss
vallerii: its a fixed width of card, wich has to be located to according to the indicated number of repetitions.

silchyn: @vallerii I mean you should use variable for that

silchyn (approved):
Cool!

PR #844, opened 2022-08-26
product cards display changed for grid

iPanda177:
- [DEMO LINK](https://iPanda177.github.io/layout_catalog_grid/)

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -1,9 +1,27 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
+  display: grid;
+  grid-template-columns: repeat(auto-fill, 200px);
+  justify-content: center;
+  gap: 48px;
 
   max-width: 944px;
   margin: 0 auto;
+
+  @media (max-width: 487px) {
+    grid-template-columns: 200px;
+  }
+
+  @media (min-width: 488px) {
+    grid-template-columns: repeat(2, 200px);
+  }
+
+  @media (min-width: 768px) {
+    grid-template-columns: repeat(3, 200px);
+  }
+
+  @media (min-width: 1024px) {
+    grid-template-columns: repeat(4, 200px);
+  }
 
   &__card {
     display: flex;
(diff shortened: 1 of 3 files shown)


src/styles/_catalog.scss
```diff
@@ -1,10 +1,28 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
+  display: grid;
+  grid-template-columns: repeat(auto-fill, 200px);
+  justify-content: center;
+  gap: 48px;
 
   max-width: 944px;
   margin: 0 auto;
```
iam-datsenko, line 8: If u set repeat (4, 200px) - ur page will be with 4 cards, so u don't need max-width and margin

src/styles/_catalog.scss
```diff
@@ -1,10 +1,28 @@
...
+    grid-template-columns: repeat(2, 200px);
+  }
+
+  @media (min-width: 768px) {
+    grid-template-columns: repeat(3, 200px);
+  }
+
+  @media (min-width: 1024px) {
+    grid-template-columns: repeat(4, 200px);
+  }
```
iam-datsenko, line 24: If u follow my suggestion above, here u can create only 1 @media - repeat (auto-fill, 200px)

iam-datsenko: And u don't need nth child, grid has done it all for u))

iam-datsenko (approved):
U can code smarter ;)

src/styles/_catalog.scss
Denchepornyuk: and you can't use margin for BEM block, remove it

src/styles/_catalog.scss
Denchepornyuk: agree, nice suggestion

Denchepornyuk (changes requested):

src/styles/_catalog.scss
iPanda177: ooops, sry, stupid mistake ;)

iPanda177:
@Denchepornyuk thanks for your suggestions) made a few fixes as well)
But I've decided to leave the media for each resolution, cuz by the task I need to change columns at a specific resolution.
Also, u said that its better practice less auto-fill (when you was reviewing @iam-datsenko code).

Denchepornyuk (approved):

PR #848, opened 2022-08-27
catalog_grid

AlexKinda:
https://AlexKinda.github.io/layout_catalog_grid/


stanislav-topikha (changes requested):
It should be centered on smaller screens 
Screenshot 2022-08-30 at 15 27 27
Screenshot 2022-08-30 at 15 27 36

PR #861, opened 2022-08-28
Catalog grid

iam-datsenko:
[DEMO LINK](https://iam-datsenko.github.io/layout_catalog_grid/)

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -1,9 +1,9 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
-
-  max-width: 944px;
-  margin: 0 auto;
+  display: grid;
+  grid-template-columns: repeat(4, 200px);
+  justify-content: center;
+  align-content: center;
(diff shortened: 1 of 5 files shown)


src/styles/_catalog.scss
```diff
@@ -16,7 +16,7 @@
     margin-bottom: 48px;
   }
 
-  &__card:not(:nth-child(4n)) {
-    margin-right: 48px;
+  @media (max-width: 1024px) {
+    grid-template-columns: repeat(auto-fit, 200px);
```
iPanda177, line 20: this is the shortest way to add a grid, but according to the task, you need the columns to change at a specific resolution. I recommend adding media queries for each resolution.

iPanda177 (approved):

src/styles/_catalog.scss
Denchepornyuk: media queries for each resolution would be better solution but this one also good 👍

Denchepornyuk (approved):

PR #869, opened 2022-08-29
Done

tatyana22:
- [DEMO LINK](https://tatyana22.github.io/layout_catalog_grid/)


src/styles/main.scss
```diff
@@ -10,5 +10,50 @@ body {
...
+@media (min-width: 768px) {
+
+  .catalog {
+    display: grid;
+    grid-template-columns: repeat(3, $cardWidth);
+    gap: 48px;
+    justify-content: center;
+    align-content: center;
+  }
+}
```
OleksandrGlulhovskiy, line 48: You need add in media only grid-template-colums, other attributes must be allways in catalog, i think it's not serious mistake, but it will be better

OleksandrGlulhovskiy (commented):
You don't need commit all your folders(.json); At all Good job))

src/styles/main.scss
tatyana22: yes it would be better! Thanks

src/styles/main.scss
```diff
@@ -10,5 +10,42 @@ body {
...
+    grid-template-columns: repeat(3, $cardWidth);
+  }
+}
+
+@media (min-width: 1024px) {
+
+  .catalog {
+    grid-template-columns: repeat(4, $cardWidth);
+  }
+}
```
MaksimusCRS, line 51: Do you need all this media queries ?

MaksimusCRS (changes requested):

src/styles/main.scss
tatyana22: one media queries deleted, everything woks. Thanks!

MaksimusCRS (approved):
GJ

PR #871, opened 2022-08-29
Catalog was made

denyskleimenov:
- [DEMO LINK](https://DenysKleimenov.github.io/layout_catalog_grid/)


src/styles/blocks/catalog.scss
```diff
@@ -0,0 +1,41 @@
+.catalog {
+  display: grid;
+
```
korniienkokarina, line 3: it might be better to include "grid-template-columns: repeat(1, auto)" here for clarity, so it would be easier to understand code logic from 1 column to 4, even though it works just fine without it

korniienkokarina (commented):
everything looks and works fine, good job :)

Andr1yK (approved):

PR #877, opened 2022-08-29
changed flex on grid and added hovers styles

andrew-satyr:
[DEMO LINK](https:/andrew-satyr.github.io/layout_catalog_grid/)

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -10,5 +10,21 @@ body {
   box-sizing: border-box;
 }
 
+@import "./utilits/variables";
+
 @import "_header.scss";
+
 @import "_catalog.scss";
+
+.catalog {
+  display: grid;
+  grid-template-columns: repeat(4, 200px);
+  gap: $gap;
+  justify-content: center;
+
+  @media (max-width: $card-width*4 + $gap*3 + 80px) {
+    & {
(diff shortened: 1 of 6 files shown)


src/styles/main.scss
```diff
@@ -10,5 +10,21 @@ body {
...
+  grid-template-columns: repeat(4, 200px);
+  gap: $gap;
+  justify-content: center;
+
+  @media (max-width: $card-width*4 + $gap*3 + 80px) {
+    & {
+      grid-template-columns: repeat(auto-fill, $card-width);
+    }
+  }
+}
```
Andrew-Skoryk, line 30: You forget to make it specific:
Media queries
There should be one column below 488px

Andrew-Skoryk (changes requested):

src/styles/catalog.scss
```diff
@@ -0,0 +1,40 @@
+.catalog {
+  display: grid;
+  grid-template-columns: repeat(4, 200px);
+  gap: $gap;
+  justify-content: center;
+
+  max-width: 944px;
```
Andr1yK, line 7: u really need this?

Andr1yK (approved):

PR #882, opened 2022-08-30
Replace flex with grid

AniaHna:
[DEMO LINK](https://AniaHna.github.io/layout_catalog_grid/)

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -12,3 +12,27 @@ body {
 
 @import "_header.scss";
 @import "_catalog.scss";
+
+.catalog {
+  display: grid;
+  grid-template-columns: repeat(1, 1fr);
+  grid-gap: 48px;
+}
+
+@media (min-width: 488px) and (max-width: 767px) {
+  .catalog {
+    grid-template-columns: repeat(2, 1fr);
+  }
+}
+
+@media (min-width: 768px) and (max-width: 1023px) {
+  .catalog {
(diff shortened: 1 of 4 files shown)


src/styles/main.scss
```diff
@@ -12,3 +12,27 @@ body {
 
 @import "_header.scss";
 @import "_catalog.scss";
+
+.catalog {
+  display: grid;
+  grid-template-columns: repeat(1, 1fr);
+  grid-gap: 48px;
+}
```
AndriyZhabynets, line 20: We need to set justify-content for 'catalog' class, for center alignment

src/styles/main.scss
```diff
@@ -12,3 +12,27 @@ body {
...
+  .catalog {
+    grid-template-columns: repeat(3, 1fr);
+  }
+}
+
+@media (min-width: 1024px) {
+  .catalog {
+    grid-template-columns: repeat(4, 1fr);
+  }
+}
```
AndriyZhabynets, line 38: using 1fr is not a good solution in this situation, it is better to use the width of the card (200px)

AndriyZhabynets (changes requested):

src/styles/main.scss
Anna-Tkachenko: agree, will be better to use card width

Anna-Tkachenko (approved):

PR #887, opened 2022-08-30
catalog-grid

GizemEce:
- [DEMO LINK](https://GizemEce.github.io/layout_catalog_grid/)

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -1,19 +1,30 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
+  display: grid;
+  justify-content: center;
 
-  max-width: 944px;
-  margin: 0 auto;
+  gap: 48px;
+
+  @media (min-width: 1024px) {
+    grid-template-columns: repeat(4, $card-width);
+  }
+
+  @media (min-width: 768px) and (max-width: 1024px) {
+    grid-template-columns: repeat(3, $card-width);
+  }
+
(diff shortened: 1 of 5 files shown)


src/styles/_catalog.scss
```diff
@@ -1,19 +1,30 @@
...
+
+  @media (min-width: 1024px) {
+    grid-template-columns: repeat(4, $card-width);
+  }
+
+  @media (min-width: 768px) and (max-width: 1024px) {
+    grid-template-columns: repeat(3, $card-width);
+  }
+
+  @media (min-width: 488px) and (max-width: 768px) {
```
Quelaan1, line 15:
```suggestion
  @media (min-width: 488px){
```

max-width can be removed as per my understanding. And also max-width should be 1px less than the  above min-width.

Quelaan1 (approved):
Good job

src/styles/_catalog.scss
VinnieJ-2k20: Use one condition at a time by putting media queries in the correct order.
Example:
default -> media (max-width: 1023px) -> media (max-width: 767px) -> media (max-width: 477px)

VinnieJ-2k20 (changes requested):
Almost done, just one fix needed.

VinnieJ-2k20 (changes requested):
I see only 3 cards in a row on large screens, but need 4.
Something is wrong with media queries.

VinnieJ-2k20 (approved):
Good job 👍

PR #888, opened 2022-08-30
v1: initial commit

igor-1987:
- [DEMO LINK](https://igor-1987.github.io/layout_catalog_grid/)


src/styles/utils/_vars.scss
```diff
@@ -0,0 +1,9 @@
+$box-shadow: 0 2px 4px rgba(0, 0, 0, 0.05);
+$font-family: "Roboto", sans-serif;
+$color-primary: #060b35;
+$color-secondary: #fff;
+$color-border: #ccc;
+$card-height: 300px;
+$card-width: 200px;
+$duration: 300ms;
+$card-gap: 48px;
```
glugovsky, line 9: Tip: some of these variables (such as `$box-shadow`, `$font-family`, `$color-border`, `$card-height` and `$card-gap`) are used only once, so you don't need those essentially. But, considering its usefulness for controls and readability, most of it can remain. The only variable I would recommend to remove is `$font-family`, it is set to the whole html, only once, and from its position in the code it is pretty obvious where to adjust your settings without a variable. But that's really up to you :)

src/styles/blocks/_catalog.scss
```diff
@@ -0,0 +1,46 @@
...
+      background-color: $color-primary;
+      color: $color-secondary;
+    }
+  }
+
+  @media (min-width: 1024px) {
+    & {
+      @include column(4, $card-width);
+    }
+  }
```
glugovsky, line 27: Tip: media queries can be nested in scss. Suggested change will work only if you move this fragment inside `.catalog`.
```suggestion
  @media (min-width: 1024px) {
      @include column(4, $card-width);
  }
```

src/styles/blocks/_catalog.scss
```diff
@@ -0,0 +1,46 @@
...
+    & {
+      @include column(2, $card-width);
+    }
+  }
+
+  @media (max-width: 487px) {
+    & {
+      @include column(1, $card-width);
+    }
+  }
```
glugovsky, line 45: Tip: you can write this `@include` simply in `.catalog` without media queries. It will be treated as default setting and will be overwritten by another queries if those are below it

glugovsky (approved):
Looks and works fine, good job :) However, I have left some tips in comments, would be great if you applied them

src/styles/utils/_vars.scss
igor-1987: Yeah, agree.

src/styles/blocks/_catalog.scss
igor-1987: > Looks and works fine, good job :) However, I have left some tips in comments, would be great if you applied them

Thank you, it makes sense. In fact, I've relized all of it reviewing your own code.

igor-1987:
> Looks and works fine, good job :) However, I have left some tips in comments, would be great if you applied them

Thank you, it makes sense. In fact, I've relized all of it reviewing your own code.

src/styles/blocks/_catalog.scss
```diff
@@ -0,0 +1,38 @@
...
+    @include grid-col(3);
+  }
+
+  @media (min-width: 488px) and (max-width: 767px) {
+    @include grid-col(2);
+  }
+
+  @media (max-width: 487px) {
+    @include grid-col(1);
+  }
```
Denchepornyuk, line 37: you can use auto-fill instead of all these media, but media also good solution

Denchepornyuk (approved):

src/styles/blocks/_catalog.scss
igor-1987: > you can use auto-fill instead of all these media, but media also good solution

Thank you, I've tried "auto-fill". Works silmilar, but looks even cleaner.

PR #891, opened 2022-08-30
Catalog-grid

hajnaloltyan:
- [DEMO LINK](https://hajnaloltyan.github.io/layout_catalog_grid/)

I used my own catalog, I hope I understood the task correctly.


src/styles/blocks/catalog.scss
```diff
@@ -0,0 +1,21 @@
...
+  display: grid;
+  grid-template-columns: repeat(4, auto);
+  justify-content: center;
+
+  padding: $paddingY $paddingX;
+  row-gap: $gapX;
+  column-gap: $gapY;
+
+  @media (max-width: 487px) {
+    grid-template-columns: repeat(1, auto);
```
fatihsamur, line 11: I think you can skip the usage of repeat for 1 and just use auto.

fatihsamur (approved):
If I understand the task correctly, we should only change the layout of the catalog and not move forward. But your code looks fine in general 👍

src/styles/blocks/catalog.scss
hajnaloltyan: if I delete the 1, it won't work

VinnieJ-2k20 (approved):
Good job 👍

src/styles/blocks/catalog.scss
fatihsamur: Did you try like this; `grid-template-columns: auto;`

PR #921, opened 2022-09-04
re-created catalog grid

Skycember:
- [DEMO LINK](https://Skycember.github.io/layout_catalog_grid/)

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -10,5 +10,6 @@ body {
   box-sizing: border-box;
 }
 
+@import "./utils/variables.scss";
 @import "_header.scss";
 @import "_catalog.scss";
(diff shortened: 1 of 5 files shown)


src/styles/main.scss
```diff
@@ -10,5 +10,6 @@ body {
   box-sizing: border-box;
 }
```
Bibicutu, line 11: I would have put this in a `styles/utils/_resets.css` file

Bibicutu:
The location of the `_catalog.scss` file is in `src/styles/blocks/` and I would remove "_" from the name, like this: `src/styles/blocks/catalog.scss`

ihor-jpeg (approved):

PR #926, opened 2022-09-06
a catalog-grid has been added

Aleksey-Aleksey:
[DEMO](https://Aleksey-Aleksey.github.io/layout_catalog_grid/)


Virass (changes requested):
Use variable for card width

PR #987, opened 2022-10-02
catalog_grid

dmitry-fesenko86:
https://dmitry-fesenko86.github.io/layout_catalog_grid/


src/styles/blocks/_catalog.scss
```diff
@@ -0,0 +1,40 @@
...
+  .catalog {
+    grid-template-columns: repeat(3, $card-width);
+  }
+}
+
+@media (min-width: 1024px) {
+  .catalog {
+    grid-template-columns: repeat(4, $card-width);
+  }
+}
```
AlisaVlasova, line 40: you can move these media queries inside the .catalog block, then you won't need to write a selector every time

AlisaVlasova (changes requested):

AlisaVlasova (approved):

PR #1029, opened 2022-10-10
Catalog grid was added

MikhailKucherina:
- [DEMO LINK](https://MikhailKucherina.github.io/layout_catalog_grid/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -15,36 +15,24 @@
   <body>
     <header class="header">
       <a href="#home" class="header__logo">
-        <img
-          src="images/logo.png"
-          alt="logo"
-        >
+        <img src="images/logo.png" alt="logo">
       </a>
       <nav class="nav">
         <ul class="nav__list">
           <li class="nav__item">
-            <a
-              class="nav__link is-active"
-              href="#"
-            >
+            <a class="nav__link is-active" href="#">
             Apple
             </a>
           </li>
 
           <li class="nav__item">
-            <a
-              class="nav__link"
-              href="#"
-            >
+            <a class="nav__link" href="#">
             Samsung
             </a>
           </li>
 
           <li class="nav__item">
-            <a
-              class="nav__link"
-              href="#"
-            >
+            <a class="nav__link" href="#">
             Smartphones
             </a>
           </li>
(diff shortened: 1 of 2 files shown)


src/index.html
```diff
@@ -3,75 +3,443 @@
...
-
+    <title>Catalog</title>
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
     <header class="header">
-      <a class="header__logo" href="">
-        <img src="./images/logo.png" alt="Moyo logo">
+      <a href="#home" class="header__logo">
+        <img
```
shtht24, line 18: no need to set them into separate strings if it'is not more than 80 sym or parametrs less than 3. chech it all the way

src/styles/blocks/header.scss
```diff
@@ -0,0 +1,16 @@
+.header {
```
shtht24, line 1: is it necessary to use all styles from layout_catalog? i thought we got a mock-up inside layout_catalog_grid and we should just make it work correctly

src/styles/blocks/catalog.scss
```diff
@@ -0,0 +1,27 @@
+.container {
+  display: grid;
+  grid-template-columns: repeat(4, $full-card-width);
+  gap: 48px;
+  justify-content: center;
+
+  width: ($full-card-width * 4) + ($card-gap * 3);
```
shtht24, line 7: we dont need to set width. it could be measured by cards' size and quantity.

src/styles/blocks/catalog.scss
```diff
@@ -0,0 +1,27 @@
...
+
+  margin: 0 auto;
+  padding-top: 50px;
+
+  @media (max-width: 1024px) {
```
shtht24, line 12: try to realize it by min-width, this way you can drop grid-template-columns: repeat(4, $full-card-width) from .container and just set in media

shtht24 (changes requested):
almost ok, but there are too much redundant in your code. you can make it easier :)

src/styles/blocks/catalog.scss
```diff
@@ -1,24 +1,22 @@
 .container {
   display: grid;
-  grid-template-columns: repeat(4, $full-card-width);
+  grid-template-columns: repeat(1, $full-card-width);
```
shtht24, line 3: i think you can delete this 'cause this is default)

shtht24 (approved):
nice 👍

denys-danyliuk (approved):

PR #1042, opened 2022-10-11
grid is done

vlad-shkv:
- [DEMO LINK](https://vlad-shkv.github.io/layout_catalog_grid/)


DanilSerzh (changes requested):
no padding on small screens

PR #1044, opened 2022-10-12
Catalog_grid

referee1988:
- [DEMO LINK](https://referee1988.github.io/layout_catalog_grid/)

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -1,22 +1,28 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
+  display: grid;
+  grid-template-columns: repeat(1, $cardWidth);
+  justify-content: center;
+  gap: 48px;
 
-  max-width: 944px;
-  margin: 0 auto;
+  @media (min-width:488px) {
+    grid-template-columns: repeat(2, $cardWidth);
+  }
+
+  @media (min-width:768px) {
+    grid-template-columns: repeat(3, $cardWidth);
+  }
+
+  @media (min-width:1024px) {
+    grid-template-columns: repeat(4, $cardWidth);
+  }
 
   &__card {
-    display: flex;
+    display: grid;
     justify-content: center;
     align-items: center;
-    width: 200px;
+    width: $cardWidth;
     height: 300px;
     border: 1px solid #ccc;
     border-radius: 5px;
-    margin-bottom: 48px;
-  }
-
-  &__card:not(:nth-child(4n)) {
-    margin-right: 48px;
   }
 }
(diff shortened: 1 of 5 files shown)


src/styles/_catalog.scss
```diff
@@ -1,22 +1,28 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
+  display: grid;
+  grid-template-columns: repeat(1, $cardWidth);
```
AlexTsimba, line 3: You shouldn't use repeat for only one column. Just keep $cardWidth

```suggestion
  grid-template-columns: $cardWidth;
```

src/styles/main.scss
```diff
@@ -10,5 +10,7 @@ body {
   box-sizing: border-box;
 }
 
+@import "./utils/variables";
```
AlexTsimba, line 13: If you will add bottom dash to all folders, it would be in same code style
```suggestion
@import "./_utils/variables";
```

AlexTsimba (changes requested):
Great job, with few small mistakes that does not affect the result

src/styles/_catalog.scss
denys-danyliuk: agree

src/styles/_catalog.scss
```diff
@@ -1,22 +1,28 @@
...
+  justify-content: center;
+  gap: 48px;
 
-  max-width: 944px;
-  margin: 0 auto;
+  @media (min-width:488px) {
```
denys-danyliuk, line 7:
```suggestion
  @media (min-width: 488px) {
```

src/styles/_catalog.scss
```diff
@@ -1,22 +1,28 @@
...
+    grid-template-columns: repeat(2, $cardWidth);
+  }
+
+  @media (min-width:768px) {
+    grid-template-columns: repeat(3, $cardWidth);
+  }
+
+  @media (min-width:1024px) {
```
denys-danyliuk, line 15: same as above

src/styles/main.scss
denys-danyliuk: good point

src/styles/_catalog.scss
```diff
@@ -1,22 +1,28 @@
...
+    grid-template-columns: repeat(4, $cardWidth);
+  }
 
   &__card {
-    display: flex;
+    display: grid;
     justify-content: center;
     align-items: center;
-    width: 200px;
+    width: $cardWidth;
```
denys-danyliuk, line 23: it's redundant, let grid to dictate width of card

denys-danyliuk (changes requested):

mykola-koval (approved):

PR #1112, opened 2022-11-08
catalog_grid_first

myroslava-kukhta:


polosanya (changes requested):
Your cards are not centered if there are less than 4 cards
Please try to fix it
image

liliia-livandovskaya (changes requested):
The previous comment is not fixed. How it should look: 
Screenshot 2022-11-09 at 12 01 49
how it looks: 
Screenshot 2022-11-09 at 12 01 42
Please fix it or ask in fe_chat if something is not clear

qoa11a (approved):

PR #1127, opened 2022-11-12
Develop

sercansimsek:


src/styles/index.scss
```diff
@@ -0,0 +1,5 @@
+@import "../styles/utils/variables";
+
+@import "../styles/page.scss";
+@import "_header.scss";
+@import "_catalog.scss";
```
nurdanokcu, line 5: My suggestion would be using same style for naming the files and path to the files as well.  
```suggestion
@import "../styles/page.scss";
@import "../styles/header.scss";
@import "../styles/catalog.scss";
```

nurdanokcu (approved):
I see this task is done neatly.  ☑️

mgrinko (approved):

PR #1136, opened 2022-11-13
Done

mmahfuzi:
- [DEMO LINK](https://mmahfuzi.github.io/layout_catalog_grid/)

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -1,22 +1,54 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
-
-  max-width: 944px;
+  display: grid;
+  gap: $gap;
+  justify-content: center;
+  align-content: center;
   margin: 0 auto;
 
   &__card {
-    display: flex;
+    display: grid;
     justify-content: center;
     align-items: center;
     width: 200px;
     height: 300px;
     border: 1px solid #ccc;
     border-radius: 5px;
-    margin-bottom: 48px;
   }
 
   &__card:not(:nth-child(4n)) {
-    margin-right: 48px;
+    // margin-right: 48px;
+  }
+}
+
+@media (min-width: 488px) {
+  .catalog {
+    grid-template-columns: 1fr 1fr;
+    grid-template-rows: 1fr 1fr;
+    width: getCatalogWidth($cardWidth, 2);
+
+    grid-template-areas: "card card"
+      "card card";
+  }
+}
+
+@media (min-width: 768px) {
+  .catalog {
+    grid-template-columns: 1fr 1fr 1fr;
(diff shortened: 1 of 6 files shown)


src/styles/_catalog.scss
```diff
@@ -1,22 +1,57 @@
...
+@media (min-width: 488px) {
+  .catalog {
+    grid-template-columns: 1fr 1fr;
+    grid-template-rows: 1fr 1fr;
+    width: getCatalogWidth($cardWidth, 2);
+
+    grid-template-areas: "card card"
+      "card card";
   }
+}
```
CharAnagn, line 35:
```suggestion
.catalog {

      /* styles */

  @media (min-width: 488px) {
      grid-template-columns: 1fr 1fr;
      grid-template-rows: 1fr 1fr;
      width: getCatalogWidth($cardWidth, 2);
  
      grid-template-areas: "card card"
        "card card";
     }
}
```
If you put your medias inside the block that they are used for, it will be easier for you to find them and more clear tell what are for.

src/styles/_catalog.scss
```diff
@@ -1,22 +1,57 @@
...
+
+@media (min-width: 768px) {
+  .catalog {
+    grid-template-columns: 1fr 1fr 1fr;
```
CharAnagn, line 39:
```suggestion
    grid-template-columns:  repeat(3, 1fr);
```
A shorthand

src/styles/_catalog.scss
```diff
@@ -1,22 +1,57 @@
...
+
+@media (min-width: 488px) {
+  .catalog {
+    grid-template-columns: 1fr 1fr;
+    grid-template-rows: 1fr 1fr;
+    width: getCatalogWidth($cardWidth, 2);
+
+    grid-template-areas: "card card"
+      "card card";
   }
```
CharAnagn, line 34: I believe there is a lot of unnecessary code. No that it doesn't work but you can do it with less. 

Also you don't have to specify the width of the catalog for each resolution.

CharAnagn (commented):
I suggest watching again the 'Placing cards using grid' video. It will help you do it with very few lines of code. 

Also if you want, you can try using @each rule for the media queries. 

All you will need is to declare a variable with the resolutions you need and the collumns for each resolutions. $var-name: ("resolution" "columns", ... ,) and iterate with the @each rule. Try using the shorthand for the grid-template-area.

mmahfuzi:
Thanks @CharAnagn for your comments. I applied them on my code :)

mmahfuzi:
> While hovering the bottom cards row - something goes wrong
> 
>  Screen.Recording.2022-11-15.at.10.37.32.mov

Hello, what is wrong?

naziamoff (approved):

PR #1159, opened 2022-11-14
updated

Grayfox147:
- [DEMO LINK](https://Grayfox147.github.io/layout_catalog_grid/)

### DIFF
--- a/src/styles/blocks/_catalog.scss
+++ b/src/styles/blocks/_catalog.scss
@@ -14,7 +14,7 @@
     height: 300px;
     border: 1px solid #ccc;
     border-radius: 5px;
-    transition-duration: $transition-duration;
+    transition-duration: $trans-speed;
     transition-timing-function: ease;
 
     &:hover {
@@ -38,8 +38,6 @@
   @media (min-width: 1024px) {
     & {
       grid-template-columns: repeat(4, $card-width);
-      min-width: ;
     }
   }
 }
-


mmahfuzi (changes requested):
Comparing to the approved ones, I see two differences in your demo link which apparently should be fixed.
1) There is an extra space between the cards in smaller screens (less than 4 cards in each row) which is due to a right margin of the cards, and it increases when the screen gets smaller. There shouldn't be any right margin for the cards in the catalog class, only the gap is enough. So in different viewports the distance between the cards remains the same.

2) The second point, which might not be very important, is the lack of the increasing in cards by 20 percent on hover.

src/styles/_catalog.scss
```diff
@@ -13,10 +14,30 @@
...
     border-radius: 5px;
-    margin-bottom: 48px;
+
+  }
+
+  @media (max-width: 488px) {
+    & {
+      grid-template-columns: 1fr;
+    }
+  }
```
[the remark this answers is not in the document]
Grayfox147: Done!

mmahfuzi (approved):
Well done! :)

src/styles/utils/variables.scss
```diff
@@ -0,0 +1,2 @@
+$card-width: 200px;
+$trans-speed: 300ms;
```
naziamoff, line 2: better to call it not the `trans-speed`, but `transition-duration` - it looks clearer this way

naziamoff (approved):
Well done 🔥 
Just one small comment for you

src/styles/utils/variables.scss
Grayfox147: Done! thank you very much Nazar!

PR #1162, opened 2022-11-14
Catalog_grid finished

RedkoSerhii:
- [DEMO LINK](https://RedkoSerhii.github.io/layout_catalog_grid/)

### DIFF
--- a/src/blocks/_catalog.scss
+++ b/src/blocks/_catalog.scss
@@ -2,7 +2,7 @@
   display: grid;
   grid-template-columns: repeat(4, $card-width);
   justify-content: center;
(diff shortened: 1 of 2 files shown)


oleksandrnikora (approved):
Good job!
There is an insignificant note:if the value is not repeated you may not be required to use a variable.

RedkoSerhii:
> Good job! There is an insignificant note:if the value is not repeated you may not be required to use a variable.

Thanks, you right, I fixed it.

oleksandrnikora (approved):

valerii-derkach (approved):

PR #1179, opened 2022-11-15
Catalog Grid

Rijo-J:
- [DEMO LINK](https://rijo-j.github.io/layout_catalog_grid/)


sercansimsek (approved):
Nicely done! But you could give one line break after each list items

src/styles/blocks/catalog.scss
```diff
@@ -0,0 +1,37 @@
...
+  }
+
+  &__card {
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    border: 1px solid #ccc;
+    border-radius: 5px;
+
+    /* stylelint-disable */
```
naziamoff, line 23: why disabled linter?

naziamoff (changes requested):
GJ overall!
But it is not really good practice to disable the linter, usually, there are better ways of solving the issue. What was the problem that you decided to bypass it with linter disabling?

src/styles/blocks/catalog.scss
Rijo-J: It wouldn't accept the indented multi-line values, I thought it would be easier to read the values like that. Will convert to 1-line! Thank you! :rocket:

naziamoff (approved):
🔥

PR #1183, opened 2022-11-15
finished catalog-grid

JzSol:
- [DEMO LINK](https://JzSol.github.io/layout_catalog_grid/)

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -1,22 +1,38 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
-
+  display: grid;
+  justify-content: center;
+  gap: 48px;
   max-width: 944px;
   margin: 0 auto;
+
+  @media (min-width: 448px) {
+    grid-template-columns: repeat(2, 300px);
+  }
+
+  @media (min-width: 768px) {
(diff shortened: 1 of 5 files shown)


src/styles/_catalog.scss
```diff
@@ -1,22 +1,38 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
-
+  display: grid;
+  justify-content: center;
+  gap: 48px;
   max-width: 944px;
```
Alireza-Mohammadhossein, line 5: you could remove max width here, because each card has 200px width

src/styles/_catalog.scss
```diff
@@ -1,22 +1,38 @@
...
   margin: 0 auto;
 
+  @media (min-width: 448px) {
+    grid-template-columns: repeat(2, 300px);
```
Alireza-Mohammadhossein, line 9: it is better to define a width variable with 200px value, and use it instead of 200 px

Alireza-Mohammadhossein (commented):
Well done!

naziamoff (approved):
Good job overall!

But I recommend you to fix the points by @Alireza-Mohammadhossein, they make sense

PR #1186, opened 2022-11-15
completed task

Alireza-Mohammadhossein:
- [DEMO LINK](https://Alireza-Mohammadhossein.github.io/layout_catalog_grid/)


src/styles/blocks/_catalog.scss
```diff
@@ -0,0 +1,33 @@
+.catalog {
+  display: grid;
+  grid-template-columns: repeat(1, $card-width);
```
jackelashkar, line 3: Good job Ali! for that line you don't have to add it because the grid will automatically fold to 1 column when it is below 488px :)

naziamoff (approved):
Well done!
But something is wrong with the page when I hover the bottom card row - so I suggest you to fix it

PR #1189, opened 2022-11-15
commit message

miftaha:
[DEMO LINK](https://miftaha.github.io/layout_catalog_grid/)


src/styles/_catalog.scss
```diff
@@ -1,14 +1,24 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
+  display: grid;
+  grid-template-columns: repeat(1,200px);
```
Grayfox147, line 3: since you use this value many times you could create a variable for it! and store it in a variables file.

src/styles/_catalog.scss
```diff
@@ -1,14 +1,24 @@
...
 
   &__card {
-    display: flex;
+    display: grid;
     justify-content: center;
-    align-items: center;
+    align-content: center;
     width: 200px;
     height: 300px;
     border: 1px solid #ccc;
```
Grayfox147, line 24: In the cheklist they ask us not to forget to add the hover properties from our previous task, I would recommend you to add it in the header and also in the cards.

src/styles/_catalog.scss
```diff
@@ -1,14 +1,24 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
+  display: grid;
+  grid-template-columns: repeat(1,200px);
+  gap: 48px;
```
Grayfox147, line 4: so then you could set in the parent of the flex some settings to center the elements.

src/styles/blocks/_catalog.scss
```diff
@@ -0,0 +1,32 @@
...
+
+  &__card {
+    @include centering;
+    width: 200px;
+    height: 300px;
+    border: 1px solid #ccc;
+    border-radius: 5px;
+    margin-bottom: 48px;
+    transition: color $transitionduration, transform $transitionduration ;
+  }
```
Alireza-Mohammadhossein, line 26: also you can use "transition: all 0.3s" for all properties

src/styles/blocks/_header.scss
```diff
@@ -27,8 +27,12 @@
...
     line-height: 60px;
+    transition: color $transitionduration;
   }
 
+  &__item:hover {
+    color: #00acdc;
+  }
   &__item:not(:last-child) {
     margin-right: 20px;
   }
```
Alireza-Mohammadhossein, line 38: instead of margin, you can use gap for grid displays

src/styles/main.scss
```diff
@@ -1,14 +1,5 @@
...
-  box-sizing: border-box;
-}
-
-@import "_header.scss";
-@import "_catalog.scss";
+@import "./blocks/pages.scss";
+@import "./utils/variables";
+@import "./utils/mixin";
+@import "./blocks/header";
+@import "./blocks/catalog";
```
Alireza-Mohammadhossein, line 5: it would be better to use one main.scss
also main.scss is better to located outside the block folder

Alireza-Mohammadhossein (commented):
Awesome Bro!

src/styles/blocks/_header.scss
naziamoff: yeah, @Alireza-Mohammadhossein is right here, gap will do better here

src/styles/blocks/_catalog.scss
naziamoff: @Alireza-Mohammadhossein actually it is considered to be a not really good practice to use `transition: all`, better to write properties explicitly

It could be done as better though:

```
transition-property: color, transform;
transition-duration: $transitionDuration;
```

naziamoff (approved):
Well done!

PR #1195, opened 2022-11-16
grid catalog create

Andrey-Ivanov999:
- [DEMO LINK](https://Andrey-Ivanov999.github.io/layout_catalog_grid/)


src/styles/_catalog.scss
```diff
@@ -1,16 +1,23 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
+  display: grid;
+  grid-template-columns: repeat(auto-fill, $heightCard);
```
Oleksandr-Filo, line 3: You don't need variable $heightCard here, you should use variable $widthCard.
```suggestion
  grid-template-columns: $widthCard;
```

src/styles/_catalog.scss
```diff
@@ -1,16 +1,23 @@
...
+  gap: 48px;
 
   max-width: 944px;
   margin: 0 auto;
```
Oleksandr-Filo, line 8: 1. no need to specify max-width;
2. you should use justify-content: center here, not margins (also you can't use margins for BEM block)

src/styles/utils/variables.scss
```diff
@@ -0,0 +1,7 @@
+$heightCard: 300px;
+$widthCard: 200px;
+$media:
+  "1px" 1,
+  "488px" 2,
+  "768px" 3,
+  "1024px"  4;
```
Oleksandr-Filo, line 7: No need to specify (min-width: 1px) in $media, you've already created grid with 1 column and in media query you should change amount of columns.
```suggestion
$media:
  "488px" 2,
  "768px" 3,
  "1024px"  4;
```

Oleksandr-Filo (changes requested):
I can't see your HTML file in pull request

src/styles/_catalog.scss
Andrey-Ivanov999: fixed

src/styles/_catalog.scss
Andrey-Ivanov999: fixed

src/styles/utils/variables.scss
Andrey-Ivanov999: fixed
Ty for rewiew:)

ivan-kadykalo (approved):
GJ🔥

PR #1196, opened 2022-11-17
done

IoghiStudio:


SalehBUD (approved):
Perfect 🏖️

naziamoff (approved):
Nice
Would be better with some hover effect though

PR #1247, opened 2022-12-07
catalog grid

RuslanLiapin:
- [DEMO LINK](https://ruslanliapin.github.io/layout_catalog_grid/)


src/styles/_catalog.scss
```diff
@@ -1,22 +1,31 @@
...
+  display: grid;
+  gap: 48px;
+  justify-content: center;
+  grid-template-columns: repeat(auto-fill,200px);
 
-  max-width: 944px;
-  margin: 0 auto;
+  @media (max-width: 488px) {
+    grid-template-columns: 200px;
+  }
```
Viktor-Kost, line 9: It's a redundant query, because you have already set grid-template-colums above 
```suggestion

```

Viktor-Kost (changes requested):

src/styles/_catalog.scss
RuslanLiapin: image
  - There should be one column below `488px`

without this query we will have this:

image

Viktor-Kost: Ok, so in this case you do not need the rule grid-template-columns: repeat(auto-fill,200px), because you overwrite it by this query

src/styles/_catalog.scss
```diff
@@ -1,22 +1,31 @@
...
+
+  @media (min-width: 488px) {
+    grid-template-columns: repeat(2,200px);
```
Viktor-Kost, line 12: And it's a good practice to create a variable for repeated values as 200px

Viktor-Kost (changes requested):

liliia-livandovskaya (approved):
Good job!

PR #1252, opened 2022-12-08
Develop

hyrych-max:
[Demo Link](https://hyrych-max.github.io/layout_catalog_grid/)


hyrych-max:
I dont know why link doesn't work, but in my computers it's work and all project is done.

I have some problems with deploying this project, maybe thats why link doesn't work.

PR #1258, opened 2022-12-09
Created catalog using grid

AnastasiiaEm:
[DEMO LINK](https://AnastasiiaEm.github.io/layout_catalog_grid/)

Did not use media, as it works as required without them.


Viktor-Kost (changes requested):
Hello!)

You did really well with using auto-fill and really closed an issue. 
But this task is not only about using grid, it's also about starting using media-queries. 
So, that's why, please rewrite with media. 

Thanks!)

Viktor-Kost (approved):
Well done!

PR #1262, opened 2022-12-10
finished

barantarasNEW:
https://barantarasNEW.github.io/layout_catalog_grid/


src/styles/_catalog.scss
```diff
@@ -1,22 +1,27 @@
-.catalog {
-  display: flex;
-  flex-wrap: wrap;
+@mixin responsive($minWidth, $howRepeat) {
+  @media (min-width: #{$minWidth}) {
+    grid-template-columns: repeat($howRepeat, minmax(200px, 1fr));
```
ivanchuk-yurii, line 3:
```suggestion
    grid-template-columns: repeat($howRepeat, 200px);
```
bc now it looks wrong, move 200px into variable
image

ivanchuk-yurii (changes requested):

src/styles/_catalog.scss
```diff
@@ -1,22 +1,27 @@
-.catalog {
-  display: flex;
-  flex-wrap: wrap;
+@mixin responsive($minWidth, $howRepeat) {
+  @media (min-width: #{$minWidth}) {
+    grid-template-columns: repeat($howRepeat, 200px);
```
ivanchuk-yurii, line 3: move `200px` into variable

ivanchuk-yurii (changes requested):
should be centered, check demo before requesting new review 
image

ivanchuk-yurii (changes requested):

src/styles/_catalog.scss
```diff
@@ -1,22 +1,30 @@
-.catalog {
-  display: flex;
-  flex-wrap: wrap;
+$widthCard: 200px;
 
+@mixin responsive($minWidth, $howRepeat) {
```
Viktor-Kost, line 3: It's not obvious to use mixins like this
I call @include responsive, ok i pass $minWidth, it's ok. 
But $howRepeat does not applies to responsive. It's about grid. 
So what if i would like to have another 5-10 different styles that have to be changed inside responsive mixin?
Please, rewrite without using mixin like this.

Viktor-Kost (changes requested):

OlEzhA0 (changes requested):
Hi! Check my comments

OlEzhA0 (approved):
Ok, I approve without the change that I left last time (try to use `@each`)
But anyway it's good practice for you to try

PR #1290, opened 2022-12-15
commit message

Peti51:


qoa11a (changes requested):
Let's add Demo Link and Test Report Link.

Peti51:
https://Peti51.github.io/layout_catalog_grid/
I only had the Demo link in the red.me file

src/styles/_catalog.scss
```diff
@@ -20,3 +20,35 @@
...
   }
 }
+
+@media (min-width: 1024px) {
+  .catalog {
+    display: grid;
+    min-width: calc(200px * 4 + 48px * 3);
+    grid-auto-flow: row;
+  }
+}
```
polosanya, line 30:
```suggestion
@media (min-width: 1024px) {
    min-width: calc(200px * 4 + 48px * 3);
}
```
Just move media into the catalog selector and don't repeat yourself.
Also, no need to duplicate the same properties all the time. Set them just for the catalog and in media set only properties that are changing.
Btw, you can use this property to handle amount of cards instead:
`grid-template-columns: repeat(2/3/4, 200px);`

src/styles/media_querise.css
```diff
@@ -0,0 +1,31 @@
+@media (min-width: 1024px) {
```
polosanya, line 1: Why do you need this file? It just duplicates the content from catalog.scss

polosanya (changes requested):
Please check the comments

src/styles/_catalog.scss
```diff
@@ -20,3 +20,27 @@
...
+@media (max-width: 767px) {
+  .catalog {
+    grid-template-columns: repeat(2, 200px);
+  }
+}
+
+@media (max-width: 487px) {
+  .catalog {
+    grid-template-columns: repeat(1, 200px);
+  }
```
polosanya, line 45: The previous comment is not fixed. Use all the advantages of SASS, and don't duplicate yourself
Also, you can optimize your code by adding default styles for the mobile version and changing the number of columns when the screen width increases.
```suggestion
grid-template-columns: repeat(1, 200px);

@media (min-width: 486px) {
    grid-template-columns: repeat(2, 200px);
}

@media (min-width: 768px) {
    grid-template-columns: repeat(3, 200px);
 }

@media (min-width: 1024px) {
    grid-template-columns: repeat(4, 200px);
  }
}
```

polosanya (changes requested):

jstmpelowycz (approved):

PR #1293, opened 2022-12-16
Added index.html, styles and readme

yurakosh701:
- [DEMO LINK](https://yurakosh701.github.io/layout_catalog_grid/)

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -1,22 +1,14 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
-
+  display: grid;
+  gap: 48px;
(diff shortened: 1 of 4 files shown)


loralevitska (changes requested):
1. fix this behaviour and check other media)
image
2. it would be good to center cards

src/styles/_catalog.scss
```diff
@@ -1,22 +1,15 @@
 .catalog {
```
polosanya, line 1: You can call the file just `catalog.scss`. No need to underline if it's a file for block

polosanya (approved):
Good job!

PR #1302, opened 2022-12-17
first

Vladqwe:
[DEMO LINK](https://Vladqwe.github.io/layout_catalog_grid/)

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -12,3 +12,31 @@ body {
 
 @import "_header.scss";
 @import "_catalog.scss";
+
+.catalog {
+  display: grid;
+  grid-template-columns: repeat(4, 1fr);
+  gap: 48px;
+
+  @media screen and (max-width: 1024px) {
+    & {
+      grid-template-columns: repeat(3, 1fr);
+    }
+  }
+
+  @media screen and (max-width: 768px) {
+    & {
+      grid-template-columns: repeat(2, 1fr);
+    }
+  }
+
+  @media screen and (max-width: 488px) {
+    & {
+      grid-template-columns: repeat(1, 1fr);
+    }
+  }
+
+  &__card {
+    width: 200px;
+  }
+}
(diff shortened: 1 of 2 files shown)


src/styles/main.scss
```diff
@@ -12,3 +12,31 @@ body {
...
+
+.catalog {
+  display: grid;
+  grid-template-columns: repeat(4, 1fr);
+  gap: 48px;
+
+  @media screen and (max-width: 1024px) {
+    & {
+      grid-template-columns: repeat(3, 1fr);
+    }
```
tetiana-ivanchenko, line 24:
```suggestion
    grid-template-columns: repeat(3, 1fr);
```

tetiana-ivanchenko: fix it everywhere

tetiana-ivanchenko: and also, try to use mobile variant first, most of users now use phones

tetiana-ivanchenko (changes requested):

tetiana-ivanchenko (commented):
image
according to design some text must be on new line

tetiana-ivanchenko (commented):
add cursor pointer here
image

fix aligning accordin to design
image

it's not according to design
image

tetiana-ivanchenko (commented):
fix it acording to design
image

tetiana-ivanchenko (changes requested):
footer is not according to design
image

tetiana-ivanchenko (changes requested):
GJ! fix your layout according to design

src/styles/main.scss
```diff
@@ -12,3 +12,25 @@ body {
 
 @import "_header.scss";
 @import "_catalog.scss";
+
+.catalog {
+  display: grid;
+  grid-template-columns: repeat(4, 1fr);
```
OlEzhA0, line 18:
```suggestion
  grid-template-columns: repeat(4, 200px);
```

OlEzhA0 (changes requested):
Screenshot 2022-12-17 at 4 47 45 PM

gap between cards more than we expecting 


it's not ok for 1025px
Screenshot 2022-12-17 at 4 48 32 PM

and check for other sizes

DarkMistyRoom (changes requested):
the catalog should be centered 
image

SanyaBratashchuk (approved):

PR #1310, opened 2022-12-17
catalog grid was implemented

OlyaZaiets:
- [DEMO LINK](https://OlyaZaiets.github.io/layout_catalog_grid/)

### DIFF
--- a/src/styles/blocks/catalog.scss
+++ b/src/styles/blocks/catalog.scss
@@ -1,0 +1,26 @@
+.catalog {
+  display: grid;
+  place-content: center;
+  grid-template-columns: $colum-width;
+  gap: 48px;
+
+  @media (min-width: 488px) {
+    grid-template-columns: repeat(2, $colum-width);
+  }
+
+  @media (min-width: 768px) {
+    grid-template-columns: repeat(3, $colum-width);
+  }
+ 
+  @media (min-width: 1024px) {
+    grid-template-columns: repeat(4, $colum-width);
+  }
+
+  &__card {
+    display: grid;
+    place-items: center;
+    height: 300px;
+    border: 1px solid #ccc;
+    border-radius: 5px;
+  }
+}
--- a/src/styles/blocks/header.scss
+++ b/src/styles/blocks/header.scss
@@ -1,0 +1,43 @@
+.header {
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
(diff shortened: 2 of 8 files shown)


src/styles/blocks/catalog.scss
```diff
@@ -0,0 +1,26 @@
...
+    grid-template-columns: repeat(2, $colum-width);
+  }
+
+  @media (min-width: 768px) {
+    grid-template-columns: repeat(3, $colum-width);
+  }
+ 
+  @media (min-width: 1024px) {
+    grid-template-columns: repeat(4, $colum-width);
+  }
```
sv-dubov, line 17: You can try to use mixin here

sv-dubov (approved):
Well done!

src/styles/blocks/header.scss
```diff
@@ -0,0 +1,43 @@
...
+  box-shadow: 0 2px 4px rgba(0, 0, 0, 0.05);
+
+  overflow: hidden;
+
+  &__logo {
+    height: 40px;
+  }
+}
+
+.nav {
```
Anna-Tkachenko, line 18: styles for `nav` bem block should be in separate scss file

src/styles/blocks/header.scss
```diff
@@ -0,0 +1,43 @@
+.header {
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+  height: 60px;
+  padding: 0 50px;
+  margin-bottom: 50px;
```
Anna-Tkachenko, line 7: you can't set margins for bem block

src/styles/utils/_variables.scss
```diff
@@ -0,0 +1 @@
+$colum-width: 200px;
```
Anna-Tkachenko, line 1: will be better to name it card width instead of the column one

Anna-Tkachenko (changes requested):

src/styles/blocks/header.scss
OlyaZaiets: According to the task, we shouldn't modify the header

src/styles/blocks/header.scss
Anna-Tkachenko: you didn't fix comment about setting margins for bem block

Anna-Tkachenko (approved):

PR #1314, opened 2022-12-18
grid was added to catalog

chiper9:
- [DEMO LINK](https://chiper9.github.io/layout_catalog_grid/)


src/styles/blocks/_catalog.scss
```diff
@@ -0,0 +1,31 @@
+.catalog {
+  display: grid;
+  grid-template-columns: 1fr;
+  grid-auto-flow: row;
```
andrii-andriushchenko, line 4: There is no nedd to use `grid-auto-flow: row;`. It is default value

src/styles/blocks/_catalog.scss
```diff
@@ -0,0 +1,31 @@
...
+  grid-gap: 48px;
+  justify-items: center;
+  align-items: center;
```
andrii-andriushchenko, line 7: I think it's useless, without `align-items: center;` everything work same

src/styles/blocks/_catalog.scss
```diff
@@ -0,0 +1,31 @@
...
+
+  &__card {
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    width: 200px;
+    height: 300px;
+    border: 1px solid #ccc;
+    border-radius: 5px;
+    margin-bottom: 48px;
```
andrii-andriushchenko, line 29: You don't have to use `margin-bottom: 48px;`, because u use `gap`

src/styles/blocks/_catalog.scss
```diff
@@ -0,0 +1,31 @@
+.catalog {
+  display: grid;
+  grid-template-columns: 1fr;
```
andrii-andriushchenko, line 3: Your `grid-column-width` is bigger than `card-width`, I think in this task we have to use the same width (card-width = column-width), and it have to work with different widths (`@media`). But in task I didn't see it requirement

andrii-andriushchenko (changes requested):
Nicely done 🚀 Here is some notes that you have to fix. But it's my opinion, and I can be wrong, so it is up to you)

chiper9:
No, you are totally right. Thank you Andrii

ArturErmolenko (approved):

PR #1316, opened 2022-12-18
replacing flex

Anton-Zhytonbaiev:
- [DEMO LINK](https://Anton-Zhytonbaiev.github.io/layout_catalog_grid/)

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -1,9 +1,29 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
-
   max-width: 944px;
   margin: 0 auto;
+
+  display: grid;
+  justify-content: center;
+  grid-template-columns: repeat(4, 200px);
+  gap: 48px;
+
+  @media (max-width: 1024px) {
+    & {
+      grid-template-columns: repeat(3, 200px);
+    }
+  }
+
+  @media (max-width: 768px) {
+    & {
+      grid-template-columns: repeat(2, 200px);
+    }
+  }
+
+  @media (max-width: 488px) {
+    & {
+      grid-template-columns: repeat(1, 200px);
+    }
+  }
 
   &__card {
     display: flex;
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -4,32 +4,26 @@
 
   display: grid;
   justify-content: center;
-  grid-template-columns: repeat(4, 200px);
+  grid-template-columns: repeat(4, $width-card);
   gap: 48px;
 
   @media (max-width: 1023px) {
-    & {
-      grid-template-columns: repeat(3, 200px);
-    }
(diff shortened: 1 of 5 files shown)


OlyaZaiets:
Did a good job, but need to do some modification. 
Doesn't develop main.scss, normalize.scss, and reset files.

src/styles/_catalog.scss
```diff
@@ -1,10 +1,30 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
-
   max-width: 944px;
   margin: 0 auto;
```
OlyaZaiets, line 3: should we save this code?

src/styles/_catalog.scss
```diff
@@ -1,10 +1,30 @@
...
 
+  display: grid;
+  justify-content: center;
+  grid-template-columns: repeat(4, 200px);
+  gap: 48px;
+
+  @media (max-width: 1024px) {
+    & {
```
OlyaZaiets, line 11: We can delete this row

src/styles/_catalog.scss
```diff
@@ -1,10 +1,30 @@
...
+    }
+  }
+
+  @media (max-width: 488px) {
+    & {
+      grid-template-columns: repeat(1, 200px);
+    }
+  }
+
   &__card {
```
OlyaZaiets, line 28: Don't need pseudo-class here

OlyaZaiets: Big distance between rows 
image

src/styles/_catalog.scss
```diff
@@ -1,10 +1,30 @@
...
   margin: 0 auto;
 
+  display: grid;
+  justify-content: center;
+  grid-template-columns: repeat(4, 200px);
+  gap: 48px;
+
+  @media (max-width: 1024px) {
+    & {
+      grid-template-columns: repeat(3, 200px);
```
OlyaZaiets, line 12: According to the task,  four columns start at 1024px, but you have 3 columns. The same problem with 768, 488, and less than 488 px
image

src/styles/_catalog.scss
```diff
@@ -1,10 +1,30 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
-
   max-width: 944px;
   margin: 0 auto;
 
+  display: grid;
+  justify-content: center;
+  grid-template-columns: repeat(4, 200px);
```
Anna-Tkachenko, line 7: create separate variable for the card width value

src/styles/_catalog.scss
Anna-Tkachenko: agree, fix @OlyaZaiets comment, this line is redundant

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
Good job!

PR #1332, opened 2022-12-20
task done

tecHamster:
- [DEMO LINK](https://tecHamster.github.io/layout_catalog_grid/)


Anastasiia-Svintsova (changes requested):
- You should use gap here instead of margin
image
- There should be 3 columns. Checklist: There should be four columns starting at 1024px. Fix all media queries according to the task, please. 
image

loralevitska (approved):
well done!

PR #1342, opened 2022-12-26
layout_catalog_grid by Intellium7

Intellium7:
[DEMO LINK](https://Intellium7.github.io/layout_catalog_grid/)

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -1,9 +1,11 @@
 $transition-duration: 300ms;
+$card-width: 200px;
+$card-height: 300px;
 
 .catalog {
   display: grid;
-  grid-template-columns: repeat(auto-fill, 200px);
-  grid-template-rows: repeat(auto-fill, minmax(200px, 4fr));
+  grid-template-columns: repeat(auto-fill, $card-width);
+  grid-template-rows: repeat(auto-fill, minmax($card-width, 4fr));
   gap: 48px;
   max-width: 944px;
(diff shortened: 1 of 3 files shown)


src/styles/_catalog.scss
```diff
@@ -1,22 +1,65 @@
...
-    margin-bottom: 48px;
-  }
 
-  &__card:not(:nth-child(4n)) {
-    margin-right: 48px;
+    transition: transform, color, $transition-duration linear;
+
+    &:hover {
+      width: 200px;
+      height: 300px;
```
ViktorHavryliuk, line 60: Don't need to repeat width and height here in hover

src/styles/_catalog.scss
```diff
@@ -1,22 +1,65 @@
...
+    & {
+      display: grid;
+      grid-template-columns: repeat(3, minmax(200px, 3fr));
+      grid-template-rows: repeat(auto-fill, minmax(200px, auto));
+    }
+  }
+
+  @media (min-width: 1024px) {
+    & {
+      display: grid;
```
ViktorHavryliuk, line 38: Don't need to set dispaly grid in media queries, it won't change from what you set in catalog class

ViktorHavryliuk (changes requested):

ViktorHavryliuk (approved):
Approved. but it's better to use variables for things like dimensions. Please add those and use them here and when setting up grid columns size
image

PR #1360, opened 2023-01-03
Develop

venher-v:
- [DEMO LINK](https://venher-v.github.io/layout_catalog_grid/)


src/styles/blocks/content.scss
```diff
@@ -0,0 +1,20 @@
...
+  @media (max-width: 1023px) {
+    grid-template-columns: repeat(3, 200px);
+  }
+
+  @media (max-width: 767px) {
+    grid-template-columns: repeat(2, 200px);
+  }
+
+  @media (max-width: 487px) {
+    grid-template-columns: 200px;
```
SanyaBratashchuk, line 18: use variable here instead of 200px

SanyaBratashchuk (changes requested):
Cards should be centered 
image

src/styles/blocks/content.scss
SanyaBratashchuk: move 200px to the variable

SanyaBratashchuk (changes requested):

PR #1362, opened 2023-01-03
Catalog grid added

VKonoplianyi:
- [DEMO LINK](https://<your_account>.github.io/layout_catalog_grid/)


DarkMistyRoom (changes requested):
check the breakpoints. There should be four columns starting at 1024px
image

same for all these queries
image

SanyaBratashchuk (approved):

PR #1367, opened 2023-01-05
Project done

dvitchenko:
- [DEMO LINK](https://dvitchenko.github.io/layout_catalog_grid/)


DarkMistyRoom (changes requested):
Improve the solution. It doesn't work in the right way
image

image

ViktorHavryliuk (changes requested):
Container is not centered, and gap between cards is too big for smaller screen sizes:
image

DarkMistyRoom (approved):
Good

PR #1372, opened 2023-01-06
layout catalog grid2

Daria13k:
https://Daria13k.github.io/layout_catalog_grid/

### DIFF
--- a/src/styles/utils/mixins.scss
+++ b/src/styles/utils/mixins.scss
@@ -1,0 +1,5 @@
+@mixin width ($m-width,$columns) {
+  @media (min-width:$m-width) {
+    grid-template-columns: repeat($columns, $card-size);
+  }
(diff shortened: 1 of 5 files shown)


src/styles/utils/mixins.scss
```diff
@@ -0,0 +1,5 @@
+@mixin width ($m-width,$columns) {
+  @media (min-width:$m-width) {
```
OlEzhA0, line 2:
```suggestion
  @media (min-width:$m-width) {
```

OlEzhA0, line 2:
```suggestion
@mixin width ($m-width, $columns) {
  @media (min-width: $m-width) {
```

OlEzhA0 (approved):
GJ!

PR #1374, opened 2023-01-06
replaced flex with grid

olya-girnyak:
- [DEMO LINK](https://olya-girnyak.github.io/layout_catalog_grid/)


Anastasiia-Svintsova (changes requested):
- There should be four columns starting at 1024px
image
- There should be three columns starting at 768px
image
- Same with these items 
image
- You already have a gap, so remove margins
image

qoa11a (approved):

PR #1405, opened 2023-01-17
Grid catalog

HannaVasylieva:
[DEMO LINK](https://HannaVasylieva.github.io/layout_catalog_grid/)


src/styles/_catalog.scss
```diff
@@ -1,22 +1,36 @@
...
-  display: flex;
-  flex-wrap: wrap;
+$width: 200px;
 
-  max-width: 944px;
-  margin: 0 auto;
+.catalog {
+  display: grid;
+  justify-content: center;
+  align-content: center;
```
witflash, line 6: Redundant rule (it's for flex only)

src/styles/_catalog.scss
```diff
@@ -1,22 +1,36 @@
-.catalog {
-  display: flex;
-  flex-wrap: wrap;
+$width: 200px;
 
-  max-width: 944px;
-  margin: 0 auto;
+.catalog {
+  display: grid;
+  justify-content: center;
```
witflash, line 5: It's the unusual usage of this rule outside of flex. Not all browsers support this approach.
Could you remove it and use `margin: auto` and `max-width` (defined for every media query)

witflash (changes requested):
Good work!
Your solution will be correct in most modern browsers but not for all versions.
Could you try to do small changes according to my comments?

HannaVasylieva:
Надеюсь все правильно поняла)

witflash (commented):
Так, все супер 👍

witflash (approved):

PR #1409, opened 2023-01-20
layout catalog grid

Viktoriia6666:
- [DEMO LINK](https://Viktoriia6666.github.io/layout_catalog_grid/)

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -1,22 +1,39 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
+  display: grid;
+  grid-template-columns: repeat(4, 200px);
+  gap: 48px;
+  justify-content: center;
 
-  max-width: 944px;
-  margin: 0 auto;
+  @media (min-width: 1024px) {
+    & {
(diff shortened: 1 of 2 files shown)


DarkMistyRoom (changes requested):
check the breakpoints. There should be three columns starting at 768px
image
image

src/styles/_catalog.scss
```diff
@@ -1,22 +1,39 @@
...
+    & {
+      grid-template-columns: repeat(2, 200px);
+    }
+  }
+
+  @media (max-width: 487px) {
+    & {
+      grid-template-columns: repeat(1, 200px);
+    }
+  }
```
loralevitska, line 29: Better start from mobile mode and move to desktop mode with  `min-width`.
Example 
- (min-width: 488px) - 2 cards
- (min-width: 768px) - 3 cards

It helps you avoid redundant code with `max-width`

loralevitska (changes requested):
Almost perfect) Let's improve

TarasHoliuk (approved):
Nicely done :heavy_check_mark:

PR #1415, opened 2023-01-21
commit message

AlexRaven93:
- [DEMO LINK](https://AlexRaven93.github.io/layout_catalog_grid/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -17,35 +17,35 @@
   </head>
   <body>
     <header class="header">
-      <a class="header__logo" href="">
+      <a class="header__logo" href="/">
         <img src="./images/logo.png" alt="Moyo logo">
       </a>
 
       <nav class="nav">
         <ul class="nav__list">
           <li class="nav__item nav__item--is-active">
-            <a href="" class="nav__link">Apple</a>
+            <a href="/" class="nav__link">Apple</a>
           </li>
           <li class="nav__item">
-            <a href="" class="nav__link">Samsung</a>
+            <a href="/" class="nav__link">Samsung</a>
           </li>
           <li class="nav__item">
-            <a href="" class="nav__link">Smartphones</a>
+            <a href="/" class="nav__link">Smartphones</a>
           </li>
           <li class="nav__item">
-            <a href="" class="nav__link">Laptops & computers</a>
+            <a href="/" class="nav__link">Laptops & computers</a>
           </li>
           <li class="nav__item">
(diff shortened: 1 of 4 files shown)


src/styles/_mixin.scss
```diff
@@ -0,0 +1,17 @@
...
+  }
+
+  @media (min-width:489px) and (max-width:768px) {
+    grid-template-columns: repeat(2, 1fr);
+  }
+
+  @media (max-width:488px) {
+    grid-template-columns: repeat(1, 1fr);
+  }
+}
```
Volodymyr-Melnyk-Fron10der, line 17: Move from mobile to desktop:

```
 grid-template-columns: repeat(1, 1fr);
 
@media (min-width:488px) {
   grid-template-columns: repeat(2, 1fr);
 }
  
@media (min-width:768px) {
   grid-template-columns: repeat(3, 1fr);
 }
  ...
```

src/styles/_catalog.scss
```diff
@@ -1,22 +1,19 @@
...
-  flex-wrap: wrap;
-
+  display: grid;
+  grid-template-columns: repeat(auto-fill, minmax(200px, 1fr));
+  gap: 48px;
   max-width: 944px;
   margin: 0 auto;
-
+  
+  @include media;
```
Volodymyr-Melnyk-Fron10der, line 8: It's bettore to move `@include` above the all styles 👆🏼

src/index.html
```diff
@@ -17,35 +17,35 @@
   </head>
   <body>
     <header class="header">
-      <a class="header__logo" href="">
+      <a class="header__logo" href="/">
```
Volodymyr-Melnyk-Fron10der, line 20: Set `#` instead of `/` , because there is 404 error after click on anchor

Volodymyr-Melnyk-Fron10der (changes requested):
Cool! Just few items here

DarkMistyRoom (changes requested):
- the cards should be centered and keep the gap on any width

image

- 
image
image

DarkMistyRoom (changes requested):
check the indent between cards. It should be always 48px
image

DarkMistyRoom (changes requested):
this problem is still here. try to limit the container width. or may be you forgot to redeploy the page? ask in the chat if you have questions
image

SanyaBratashchuk (approved):

PR #1418, opened 2023-01-22
task done

alena-karetnikova:
https://alena-karetnikova.github.io/layout_catalog_grid/


DarkMistyRoom (changes requested):
- the actual indent between cards should be 48px

image

- check the brakepoints
image

image
image

alena-karetnikova:
hi, gaps are already 48px , do you mean to make  in grid-template-columns only 200px, not from 200px to 1fr? 

media requires are also okay, I have double checked 
Screenshot 2023-01-23 at 16 08 44
Screenshot 2023-01-23 at 16 08 32

DarkMistyRoom (changes requested):
the `gap `property is 48, but the actual indents between cards are bigger

> do you mean to make in grid-template-columns only 200px, not from 200px to 1fr?
yes, it should help

try to check breakpoints on the deployed page. removing the max-width query may help


and better write in the chat, if you have questions

loralevitska (approved):

PR #1420, opened 2023-01-23
catalog grid

YelyzavetaMolochynska:
- [DEMO LINK](https://YelyzavetaMolochynska.github.io/layout_catalog_grid/)

### DIFF
--- a/src/styles/_mixins.scss
+++ b/src/styles/_mixins.scss
@@ -1,0 +1,6 @@
+@mixin viewport ($size) {
+  @media (max-width:#{$size}) {
+    @content;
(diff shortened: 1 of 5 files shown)


src/styles/_mixins.scss
```diff
@@ -0,0 +1,6 @@
+@mixin viewport ($size) {
+  @media (max-width:#{$size}) {
```
Viktor-Kost, line 2: it is ok, but nowadays it is better to use Mobile First approach

Viktor-Kost (approved):

PR #1449, opened 2023-02-02
catalog grid was created

KuharchukDmytro:
https://KuharchukDmytro.github.io/layout_catalog_grid/


src/styles/blocks/_catalog.scss
```diff
@@ -0,0 +1,34 @@
+.catalog {
+  &__wrapper {
```
Yekatery, line 2: Why did you add extra wrapper? You can wrote all this styles inside .catalog

Yekatery (approved):
Everything works like it should. Good job!

src/styles/blocks/_catalog.scss
OleksiiBlezniuk: Agree, this wrapper can be removed

src/styles/blocks/_catalog.scss
```diff
@@ -0,0 +1,34 @@
...
+    @media (max-width: 488px) {
+      grid-template-columns: $column-width;
+    }
+  }
+
+  &__card {
+    display: grid;
+    justify-content: center;
+    align-items: center;
+    width: 200px;
```
OleksiiBlezniuk, line 28: No need to set width here, 'cause your card should take column width

OleksiiBlezniuk, line 28: No need to set width here, 'cause your card should take column width

OleksiiBlezniuk, line 28: No need to set width here, 'cause your card should take column width

src/styles/blocks/_catalog.scss
```diff
@@ -0,0 +1,34 @@
...
+      grid-template-columns: repeat(3, $column-width);
+    }
+
+    @media (max-width: 768px) {
+      grid-template-columns: repeat(2, $column-width);
+    }
+
+    @media (max-width: 488px) {
+      grid-template-columns: $column-width;
+    }
```
OleksiiBlezniuk, line 21: Usually we use mixins for particular breakpoints instead of using media query every time

OleksiiBlezniuk (changes requested):

src/styles/blocks/_catalog.scss
KuharchukDmytro: i need this wrapper, because i set margin, and i cant set it for block

sergii-nosachenko: Unnecessary wrapper.

sergii-nosachenko (changes requested):

sergii-nosachenko (approved):
Good work 👍

PR #1461, opened 2023-02-05
Develop

kuzmi491:
https://kuzmi491.github.io/layout_catalog_grid/


src/styles/_catalog.scss
```diff
@@ -1,22 +1,27 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
+  display: grid;
+  gap: 48px;
+  justify-content: center;
 
-  max-width: 944px;
-  margin: 0 auto;
+  @media (min-width: 488px) {
```
BozhenaBodnar, line 6: you can use sass @each or maybe function to do code smoller

BozhenaBodnar (approved):
nice job:)

src/styles/_catalog.scss
OleksiiBlezniuk: Agree, loop can be useful here. And usually we using mixins for typical breakpoints

OleksiiBlezniuk (approved):

PR #1475, opened 2023-02-05
Added major part

Yevhenii0536:
- [DEMO LINK](https://yevhenii0536.github.io/layout_catalog_grid/)

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -1,16 +1,38 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
-
+  display: grid;
+  grid-template-columns: repeat(4, $card-width);
+  grid-template-rows: repeat(2, $card-height);
+  gap: 48px;
   max-width: 944px;
   margin: 0 auto;
 
+  @media (max-width: 1024px) {
+    & {
+      grid-template-columns: repeat(3, $card-width);
+      grid-template-rows: repeat(auto-fill, $card-height);
+    }
+  }
+
+  @media (min-width: 488px) and (max-width: 767px) {
+    & {
+      grid-template-columns: repeat(2, $card-width);
+      grid-template-rows: repeat(auto-fill, $card-height);
+    }
+  }
+
+  @media (max-width: 487px) {
+    & {
+      grid-template-columns: repeat(1, $card-width);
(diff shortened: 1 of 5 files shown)


ksavelyeva (approved):
Good job!

src/styles/_catalog.scss
```diff
@@ -1,16 +1,38 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
-
+  display: grid;
+  grid-template-columns: repeat(4, $card-width);
+  grid-template-rows: repeat(2, $card-height);
```
OleksiiBlezniuk, line 4: this rule can be skipped, your cards will take as much rows as they need

src/styles/_catalog.scss
```diff
@@ -1,16 +1,38 @@
...
+  gap: 48px;
   max-width: 944px;
```
OleksiiBlezniuk, line 6: This max-width can be removed

src/styles/_catalog.scss
```diff
@@ -1,16 +1,38 @@
...
   margin: 0 auto;
 
+  @media (max-width: 1024px) {
+    & {
```
OleksiiBlezniuk, line 10: Wrong usage of `&` in sass (it will work as well without `&`)

src/styles/_catalog.scss
OleksiiBlezniuk: Let's remove everywhere

src/styles/_mixins.scss
```diff
@@ -0,0 +1,4 @@
+@mixin card-size {
+  height: $card-height;
+  width: $card-width;
```
OleksiiBlezniuk, line 3: Actually there is no need to set width to card, it's enough to set it only to column, card should take all column width

OleksiiBlezniuk (changes requested):

src/styles/main.scss
```diff
@@ -10,5 +10,8 @@ body {
   box-sizing: border-box;
 }
```
sergii-nosachenko, line 11: Extract reset and normalize rules to separate files.

sergii-nosachenko (changes requested):

Andr1yK (approved):

PR #1484, opened 2023-02-05
Develop

Luba-Bykova:
-[DEMO LINK](https://Luba-Bykova.github.io/layout_catalog_grid/)


ViktorHavryliuk (changes requested):
Gap between cards must be 48px, container with cards grid should be centered
image
Also, why do you have two _catalog.scss files?

loralevitska (approved):
You may use only `min-width` in this cases for media

PR #1504, opened 2023-02-06
grid catalog

abcdesign1:
- [DEMO LINK](https://vadym4che.github.io/layout_catalog_grid/)

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -4,17 +4,10 @@
   align-content: center;
   gap: 48px;
 
-  @media (max-width: 487px) {
-    grid-template-columns: 200px;
-  }
-  @media (min-width: 488px) and (max-width: 767px) {
-    grid-template-columns: 200px 200px;
-  }
-  @media (min-width: 768px) and (max-width: 1023px) {
-    grid-template-columns: repeat(3, 200px);
-  }
-  @media (min-width: 1024px) {
-    grid-template-columns: repeat(4, 200px);
+  @each $columns, $breakpoint in $breakpoints {
+    @media (min-width: #{$breakpoint}) {
+      grid-template-columns:  repeat(#{$columns}, #{$card-width});
+    }
   }
 
   &__card {
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -1,9 +1,21 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
+  display: grid;
+  justify-content: center;
+  align-content: center;
+  gap: 48px;
 
-  max-width: 944px;
-  margin: 0 auto;
+  @media (max-width: 487px) {
+    grid-template-columns: 200px;
+  }
+  @media (min-width: 488px) and (max-width: 767px) {
+    grid-template-columns: 200px 200px;
+  }
+  @media (min-width: 768px) and (max-width: 1023px) {
+    grid-template-columns: repeat(3, 200px);
+  }
+  @media (min-width: 1024px) {
+    grid-template-columns: repeat(4, 200px);
+  }
 
   &__card {
     display: flex;
@@ -13,10 +25,5 @@
     height: 300px;
     border: 1px solid #ccc;
     border-radius: 5px;
-    margin-bottom: 48px;
-  }
-
-  &__card:not(:nth-child(4n)) {
-    margin-right: 48px;
   }
 }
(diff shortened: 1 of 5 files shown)


src/styles/_catalog.scss
```diff
@@ -1,9 +1,21 @@
...
+    grid-template-columns: 200px 200px;
+  }
+  @media (min-width: 768px) and (max-width: 1023px) {
+    grid-template-columns: repeat(3, 200px);
+  }
+  @media (min-width: 1024px) {
+    grid-template-columns: repeat(4, 200px);
+  }
 
   &__card {
```
sergii-nosachenko, line 20: Add width to variables

src/styles/_catalog.scss
```diff
@@ -1,9 +1,21 @@
...
+  align-content: center;
+  gap: 48px;
 
-  max-width: 944px;
-  margin: 0 auto;
+  @media (max-width: 487px) {
+    grid-template-columns: 200px;
+  }
+  @media (min-width: 488px) and (max-width: 767px) {
+    grid-template-columns: 200px 200px;
```
sergii-nosachenko, line 11: Use repeat

src/styles/_catalog.scss
```diff
@@ -1,9 +1,21 @@
...
+  justify-content: center;
+  align-content: center;
+  gap: 48px;
 
-  max-width: 944px;
-  margin: 0 auto;
+  @media (max-width: 487px) {
+    grid-template-columns: 200px;
+  }
+  @media (min-width: 488px) and (max-width: 767px) {
```
sergii-nosachenko, line 10: You can keep only min-width options. Don't need to combine it with max-width.

src/styles/_catalog.scss
```diff
@@ -1,9 +1,21 @@
...
+    grid-template-columns: 200px 200px;
+  }
+  @media (min-width: 768px) and (max-width: 1023px) {
+    grid-template-columns: repeat(3, 200px);
+  }
+  @media (min-width: 1024px) {
+    grid-template-columns: repeat(4, 200px);
+  }
```
sergii-nosachenko, line 18: You can make a loop to create all media queries. Just create a correctly mapped list to iterate.

src/styles/_catalog.scss
```diff
@@ -1,9 +1,21 @@
...
+  display: grid;
+  justify-content: center;
+  align-content: center;
+  gap: 48px;
 
-  max-width: 944px;
-  margin: 0 auto;
+  @media (max-width: 487px) {
+    grid-template-columns: 200px;
+  }
```
sergii-nosachenko, line 9: No need to create a query for the default state.

sergii-nosachenko (changes requested):

src/styles/_catalog.scss
vadym4che: image

src/styles/_catalog.scss
vadym4che: maybe yes but
if I'll do the way you suggesting me
i'll have a strong need to watch out that next resolution is going exactly next in cascading style sheet file

do you really think I actually have to do all as short as it possible with risk of broking styles scenario?

src/styles/_catalog.scss
vadym4che: It only will be shortened with previously suggested avoiding max-width I think

src/styles/_catalog.scss
vadym4che: I'll make complete investigation according to this point and surely give a feedback about this point

src/styles/_catalog.scss
vadym4che: separated all with  $breakpoints collection

src/styles/_catalog.scss
vadym4che: @each scss loop applied

src/styles/_catalog.scss
vadym4che: max-width conditional statements had been removed because of useless

src/styles/_catalog.scss
vadym4che: got it - have used it - following the best practices

src/styles/_catalog.scss
vadym4che: unnecessary query states has been removed

illiarerun (approved):

PR #1514, opened 2023-02-08
Undertake layout catalog grid task

Volodymyr-Terzov:
DEMO LINK - https://Volodymyr-Terzov.github.io/layout_catalog_grid/


src/styles/utils/_media.scss
```diff
@@ -0,0 +1 @@
+
```
DarkMistyRoom, line 1: may be you can remove this file?)

src/styles/blocks/catalog.scss
```diff
@@ -0,0 +1,38 @@
+@import "../utils/variables";
+
+.catalog {
+  display: grid;
+  grid-template-columns: repeat(auto-fit, $catalog-card-width);
+  gap: 48px;
+
+  max-width: 944px;
+  margin: 0 auto;
```
DarkMistyRoom, line 9: Don't add external styles (positioning or margins) to BEM-blocks. Use mix where necessary and move all external styles under element selector.

DarkMistyRoom (changes requested):
the catalog should be centered on any width
image

src/styles/blocks/catalog.scss
Volodymyr-Terzov: I've created an element selector (wrapper), and have moved it to a separate file and separate folder. Though, not sure whether it is in line with BEM...

PR #1530, opened 2023-02-14
first commit  job done

RatiBorRozGa:
- [DEMO LINK](https://@RatiBorRozGa.github.io/layout_catalog_grid/)


src/styles/_catalog.scss
```diff
@@ -1,6 +1,24 @@
...
-  display: flex;
-  flex-wrap: wrap;
+  display: grid;
+  gap: 48px;
+
+  grid-template-rows: repeat(auto-fill, 1fr);
+
+  @media (max-width:487px) {
+    grid-template-columns: repeat(1, 1fr);
+  }
```
DarkMistyRoom, line 9:
```suggestion
  grid-template-rows: $card-width;
```

DarkMistyRoom: use default value for less than 488px. and use hardcoded size instead of 1fr

DarkMistyRoom (changes requested):
the distance between cards should be always 48px. and they always should be centered
image

tetiana-ivanchenko (approved):

PR #1550, opened 2023-02-21
cataloggrid0.1

OleksiiBielovolov:
- [DEMO LINK](https://OleksiiBielovolov.github.io/layout_catalog_grid/)

### DIFF
--- a/src/styles/utilities/_mixins.scss
+++ b/src/styles/utilities/_mixins.scss
@@ -1,0 +1,91 @@
+@mixin oneColumn {
+  @media (min-width: 0) and (max-width: 488px) {
+    .catalog {
+      display: grid;
+      justify-content: center;
+      max-width: 944px;
+      margin: 0 auto;
+      gap: 48px;
+      grid-template-columns: 1;
+
+      &__card {
+        display: grid;
+        justify-content: center;
+        align-items: center;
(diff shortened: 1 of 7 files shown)


src/styles/utilities/_mixins.scss
```diff
@@ -0,0 +1,91 @@
...
+    }
+  }
+}
+
+@mixin twoColumns {
+  @media (min-width: 488px) and (max-width: 768px) {
+    .catalog {
+      display: grid;
+      justify-content: center;
+      max-width: 944px;
```
loralevitska, line 29: An interesting approach to solving the problem, but your code is duplicated and if the designer changes the data, it will have to be changed in all places. Rewrite the code so that your mixins are inside classes. Fit it everywhere
Example: 
```
.myButton {
  color: blue;
  margin-top: 8px;
  padding: 8px;

   @media (min-width: 488px) {
    color: red;
  }

  @media (min-width: 768px) {
    color: black;
  }

  @media (min-width: 1024px) {
    color: gray;
    padding: 12px;
  }
}
```

loralevitska (changes requested):
Good job! There are things that should be fixed:

ViktorHavryliuk (approved):

PR #1554, opened 2023-02-22
changed flex to grid

vpdrabynko:
https://vpdrabynko.github.io/layout_catalog_grid/

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -1,22 +1,34 @@
-.catalog {
-  display: flex;
-  flex-wrap: wrap;
+$card-width: 200px;
 
+.catalog {
+  display: grid;
+  gap: 48px;
   max-width: 944px;
   margin: 0 auto;
+
+  @media (max-width: 488px) {
+    grid-template-columns: repeat(1, $card-width);
+  }
+
+  @media (min-width: 488px) {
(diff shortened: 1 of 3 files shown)


src/styles/_catalog.scss
```diff
@@ -1,22 +1,34 @@
...
 
+.catalog {
+  display: grid;
+  gap: 48px;
   max-width: 944px;
   margin: 0 auto;
 
+  @media (max-width: 488px) {
+    grid-template-columns: repeat(1, $card-width);
+  }
```
Anastasiia-Svintsova, line 11: Don't think that it's useful here. You can remove this media query

src/styles/_catalog.scss
```diff
@@ -1,22 +1,34 @@
-.catalog {
-  display: flex;
-  flex-wrap: wrap;
+$card-width: 200px;
 
+.catalog {
+  display: grid;
+  gap: 48px;
   max-width: 944px;
```
Anastasiia-Svintsova, line 6: Don't hardcode the width here. You have columns already

Anastasiia-Svintsova (changes requested):
- Would be good to align the container to the center. It doesn't look good now. But cards still should start from the left. Just align the container

anstsot (approved):
Great job, approved!

PR #1579, opened 2023-03-02
added catalog using grid istead flex

PavloReutskiy:
https://PavloReutskiy.github.io/layout_catalog_grid/

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -1,22 +1,29 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
+  display: grid;
+  grid-template-columns: repeat(4, 200px);
+  grid-template-rows: repeat(auto-fill, 300px);
+  grid-auto-rows: 300px;
+  justify-content: center;
+  gap: 48px;
+  margin-bottom: 48px;
 
-  max-width: 944px;
-  margin: 0 auto;
+  @media (max-width: 1023px) {
+    grid-template-columns: repeat(3, 200px);
+  }
+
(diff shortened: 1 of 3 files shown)


src/styles/_catalog.scss
```diff
@@ -1,22 +1,29 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
+  display: grid;
+  grid-template-columns: repeat(4, 200px);
+  grid-template-rows: repeat(auto-fill, 300px);
```
Anastasia2403, line 4: Four colums should start at 1024px not 1023px! (see the screnshoot). And I see the same situation for all cases.
Maybe you will come up with a better idea how to do it correct)
https://prnt.sc/Z4QRtW4pwEh_

Anastasia2403 (changes requested):
It looks good but you need to fix your transition points in media queries!

src/styles/_catalog.scss
```diff
@@ -1,22 +1,29 @@
...
+  grid-auto-rows: 300px;
+  justify-content: center;
+  gap: 48px;
+  margin-bottom: 48px;
```
SerhiiKirik, line 8: It's extra
```suggestion
```

src/styles/_catalog.scss
SerhiiKirik, line 4: will be better, if you create variables for width & height of card

SerhiiKirik (approved):
Great 🔥

PavloReutskiy:
all done!

PR #1580, opened 2023-03-02
Created catalog using grid

nugardesq:
https://nugardesq.github.io/layout_catalog_grid/

### DIFF
--- a/src/styles/media.scss
+++ b/src/styles/media.scss
@@ -1,0 +1,17 @@
+@media screen and (max-width: 1024px) {
+  .catalog {
+    grid-template-columns: repeat(3, minmax(200px, 1fr));
+  }
+}
+
+@media screen and (max-width: 768px) {
+  .catalog {
+    grid-template-columns: repeat(2, minmax(200px, 1fr));
+  }
+}
+
+@media screen and (max-width: 488px) {
+  .catalog {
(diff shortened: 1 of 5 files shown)


src/styles/media.scss
```diff
@@ -0,0 +1,17 @@
+@media screen and (max-width: 1024px) {
```
anstsot, line 1: there is no need to create a separate file for these styles. it is still related to `.catalog` so please move them to `_catalog.scss` file

src/styles/media.scss
```diff
@@ -0,0 +1,17 @@
...
+  .catalog {
+    grid-template-columns: repeat(3, minmax(200px, 1fr));
+  }
+}
+
+@media screen and (max-width: 768px) {
```
anstsot, line 7: according to the task, there should be three columns starting at 768px. so you have to change value:
```suggestion
@media screen and (max-width: 767px) {
```

anstsot: please check other places as well

anstsot (changes requested):
Good job, but there are few fixes to improve your code.

The gap between cards seems to be more than 48px. Set `max-width` to the container or place cards in the middle.
image

anstsot (approved):
Great job!

PR #1590, opened 2023-03-02
Grid instead of flex

vksazonov:
- [DEMO LINK](https://vksazonov.github.io/layout_catalog_grid/)

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -1,22 +1,39 @@
-.catalog {
-  display: flex;
-  flex-wrap: wrap;
+$card-width: 200px;
+$card-gap: 48px;
 
-  max-width: 944px;
+.catalog {
+  display: grid;
+  gap: $card-gap;
+  max-width: $card-width * 4 + $card-gap * 3;
   margin: 0 auto;
+  grid-template-columns: repeat(4, 1fr);
+
+  @media (max-width: 1023px) {
+    & {
+      grid-template-columns: repeat(3, 1fr);
+    }
+  }
+
+  @media (max-width: 767px) {
+    & {
+      grid-template-columns: repeat(2, 1fr);
+    }
+  }
+
+  @media (max-width: 487px) {
+    & {
+      grid-template-columns: repeat(1, 1fr);
+    }
+  }
 
   &__card {
-    display: flex;
+    display: grid;
     justify-content: center;
     align-items: center;
-    width: 200px;
+    width: $card-width;
     height: 300px;
     border: 1px solid #ccc;
     border-radius: 5px;
-    margin-bottom: 48px;
-  }
-
-  &__card:not(:nth-child(4n)) {
-    margin-right: 48px;
+    
   }
 }
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -1,6 +1,3 @@
-$card-width: 200px;
-$card-gap: 48px;
-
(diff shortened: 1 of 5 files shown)


src/styles/_catalog.scss
```diff
@@ -1,22 +1,37 @@
...
+    }
+  }
 
   &__card {
-    display: flex;
-    justify-content: center;
-    align-items: center;
-    width: 200px;
+    display: grid;
+    justify-items: center;
```
iHateAntichrist, line 30: "Product card" must be in the center of card. You should not delete align items

kyserf (approved):
Everything is correct, but fix position of text inside the card

src/styles/_catalog.scss
```diff
@@ -1,22 +1,37 @@
...
+  @media (max-width: 487px) {
+    & {
+      grid-template-columns: repeat(1, 200px);
+    }
+  }
 
   &__card {
-    display: flex;
-    justify-content: center;
-    align-items: center;
```
jstmpelowycz, line 11: put flexbox back, you don't need to apply grid to cards

src/styles/_catalog.scss
```diff
@@ -1,22 +1,37 @@
...
-  flex-wrap: wrap;
+  display: grid;
+  gap: $card-gap;
+  justify-content: center;
+  grid-template-columns: repeat(4, 200px);
+
+  @media (max-width: 1023px) {
+    & {
+      grid-template-columns: repeat(3, 200px);
+    }
```
jstmpelowycz, line 13:
```suggestion
    grid-template-columns: repeat(3, 200px);
```

src/styles/_catalog.scss
```diff
@@ -1,22 +1,37 @@
...
+
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
+  display: grid;
+  gap: $card-gap;
+  justify-content: center;
+  grid-template-columns: repeat(4, 200px);
+
+  @media (max-width: 1023px) {
```
jstmpelowycz, line 10: you were told to use 1024px, 768px, 488px — so let's apply these numbers (neither less, nor greater)

src/styles/_catalog.scss
```diff
@@ -1,22 +1,37 @@
...
+    & {
+      grid-template-columns: repeat(2, 200px);
+    }
+  }
 
-  max-width: 944px;
-  margin: 0 auto;
+  @media (max-width: 487px) {
+    & {
+      grid-template-columns: repeat(1, 200px);
```
jstmpelowycz, line 24: you're not gonna have to keep this line

jstmpelowycz (changes requested):

src/styles/_catalog.scss
vksazonov: Vary strange, I don't have those properties

src/styles/_catalog.scss
```diff
@@ -1,22 +1,32 @@
+$card-width: 200px;
+$card-gap: 48px;
```
liliia-livandovskaya, line 2: it's better to move it to the `variables` file

liliia-livandovskaya (approved):
Great! 👍

src/styles/_catalog.scss
vksazonov: Yeah, I've got a question about that. If we have even one or two variables, should we make another file for variables?

PR #1594, opened 2023-03-02
catalog reworked

Kavon-Skela:
[DEMO LINK](https://Kavon-Skela.github.io/layout_catalog_grid/)

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -1,9 +1,22 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
+  display: grid;
+  grid-template-columns: repeat(4, 200px);
+  gap: 48px;
 
   max-width: 944px;
   margin: 0 auto;
+
+  @media (max-width: 1023px) {
+    grid-template-columns: repeat(3, 200px);
+  }
+
+  @media (max-width: 767px) {
+    grid-template-columns: repeat(2, 200px);
+  }
+
+  @media (min-width: 0) and (max-width: 487px) {
+    grid-template-columns: repeat(1, 200px);
+  }
 
   &__card {
     display: flex;
@@ -13,10 +26,5 @@
     height: 300px;
     border: 1px solid #ccc;
     border-radius: 5px;
-    margin-bottom: 48px;
-  }
-
-  &__card:not(:nth-child(4n)) {
-    margin-right: 48px;
   }
 }
--- a/readme.md
+++ b/readme.md
@@ -1,6 +1,6 @@
 # Frontend practice with catalog page using grid
 Replace `<your_account>` with your Github username and copy the links to Pull Request description:
-- [DEMO LINK](https://<your_account>.github.io/layout_catalog_grid/)
+- [DEMO LINK](https://Kavon-Skela.github.io/layout_catalog_grid/)
 
 > Follow [this instructions](https://github.com/mate-academy/layout_task-guideline#how-to-solve-the-layout-tasks-on-github)
 


src/styles/_catalog.scss
```diff
@@ -1,22 +1,32 @@
...
+
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
+  display: grid;
+  grid-template-columns: repeat(4, $card-width);
+  gap: 48px;
 
   max-width: 944px;
   margin: 0 auto;
```
Rage1502, line 9: I think that there should not be "max width" here. Since the width is set by the elements. And the grid has its own option to center the content through justify-content. Then you can get rid of the margin.

src/styles/_catalog.scss
```diff
@@ -1,22 +1,32 @@
...
+    grid-template-columns: repeat(1, $card-width);
+  }
+
   &__card {
-    display: flex;
+    display: grid;
     justify-content: center;
     align-items: center;
-    width: 200px;
+    width: $card-width;
```
Rage1502, line 27: You have already set the width of the cell in the grid container itself. You can get rid of it in the element itself

Rage1502 (commented):
Nice work, but needs to be changed.

Rage1502 (changes requested):
Nice work, but needs to be changed.

src/styles/_catalog.scss
Kavon-Skela: Max-width helps to set the position of a catalog. Without max-width and margin, it will stick to the left side of the screen

src/styles/_catalog.scss
```diff
@@ -1,22 +1,31 @@
+$card-width: 200px;
+
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
+  display: grid;
+  grid-template-columns: repeat(4, $card-width);
+  gap: 48px;
 
   max-width: 944px;
```
jstmpelowycz, line 8: you don't need it

src/styles/_catalog.scss
```diff
@@ -1,22 +1,31 @@
...
+    grid-template-columns: repeat(3, $card-width);
+  }
+
+  @media (max-width: 767px) {
+    grid-template-columns: repeat(2, $card-width);
+  }
+
+  @media (min-width: 0) and (max-width: 487px) {
+    grid-template-columns: repeat(1, $card-width);
+  }
```
jstmpelowycz, line 21: you were told to use 1024px, 768px, 488px — so let's apply these numbers (neither less, nor greater)

src/styles/_catalog.scss
```diff
@@ -1,22 +1,31 @@
...
 
+  @media (max-width: 1023px) {
+    grid-template-columns: repeat(3, $card-width);
+  }
+
+  @media (max-width: 767px) {
+    grid-template-columns: repeat(2, $card-width);
+  }
+
+  @media (min-width: 0) and (max-width: 487px) {
```
jstmpelowycz, line 19: `and` is redundant

src/styles/_catalog.scss
```diff
@@ -1,22 +1,31 @@
...
+    grid-template-columns: repeat(1, $card-width);
+  }
+
   &__card {
-    display: flex;
+    display: grid;
```
jstmpelowycz, line 24: you don't need to apply grid here

src/styles/_catalog.scss
```diff
@@ -1,22 +1,31 @@
...
+  @media (max-width: 1023px) {
+    grid-template-columns: repeat(3, $card-width);
+  }
+
+  @media (max-width: 767px) {
+    grid-template-columns: repeat(2, $card-width);
+  }
+
+  @media (min-width: 0) and (max-width: 487px) {
+    grid-template-columns: repeat(1, $card-width);
```
jstmpelowycz, line 20: you don't have to describe 1-column behavior, just 2, 3, 4

jstmpelowycz (changes requested):
there is a simpler way 🙂

src/styles/_catalog.scss
Kavon-Skela: image
image
image

but if I set exactly 1024, 768, and 488 it breaks task requirements. Checklist said that 4 columns should start at 1024px. Will it be okay, if my columns start at 489, 769, and 1025?

src/styles/_catalog.scss
```diff
@@ -1,22 +1,30 @@
+$card-width: 200px;
```
liliia-livandovskaya, line 1: its better to move it to the `variables` file

liliia-livandovskaya (approved):
Great job! 👍

PR #1602, opened 2023-03-02
change d-flex to d-grid and add responsive

Valerii05:


src/styles/_catalog.scss
```diff
@@ -20,3 +20,36 @@
     margin-right: 48px;
   }
 }
+
+@media screen and (max-width: 1024px) {
+  .catalog {
+    grid-template-columns: 1fr 1fr 1fr;
+    grid-template-rows: 1fr 1fr 1fr;
```
vlad-demchuk, line 27: why do you need `grid-template-rows`?

vlad-demchuk: change columns like that
`grid-template-columns: repeat(2, $card-width);`

vlad-demchuk: fix all such cases

vlad-demchuk (changes requested):

src/styles/_catalog.scss
```diff
@@ -1,15 +1,15 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
-
+  display: grid;
+  grid-template-columns: repeat(4, $card-width);
+  justify-content: space-between;
```
YegorKorenkov, line 4: Please use center

src/styles/_catalog.scss
```diff
@@ -20,3 +20,33 @@
     margin-right: 48px;
```
YegorKorenkov, line 20: Also, please use property gap instead of margin

YegorKorenkov (changes requested):

nazarmatsevych (approved):

PR #1618, opened 2023-03-05
griddooo

makandriival:
- [DEMO LINK](https://makandriival.github.io/layout_catalog_grid/)


liliia-livandovskaya (changes requested):
Block with cards should be centered
Screenshot 2023-03-10 at 14 54 28

jstmpelowycz (approved):

PR #1629, opened 2023-03-06
grid catalog v1

edgarlokys:
https://edgarlokys.github.io/layout_catalog_grid/


src/styles/utils/_media.scss
```diff
@@ -0,0 +1,17 @@
...
+  .catalog {
+    grid-template-columns: repeat(2, $card-width);
+  }
+}
+
+@media (max-width: $min-size-media) {
+  .catalog {
+    grid-template-columns: repeat(1, $card-width);
+  }
+}
```
IvanFesenko, line 17: It's bad practice use separate file with media while using sass. Move it inside catalog selector

IvanFesenko (changes requested):
Good job. Check my comment

src/styles/blocks/catalog.scss
```diff
@@ -0,0 +1,37 @@
+@import "./../utils/mixin";
+@import "./../utils/variables";
+
+.catalog {
```
tetiana-ivanchenko, line 4: there is example from official documentstion
media query should be inside selector's block
image

tetiana-ivanchenko (changes requested):

src/styles/blocks/catalog.scss
edgarlokys: i think fixed it

edgarlokys: check plz 'correct v2'

src/styles/utils/_media.scss
```diff
@@ -0,0 +1,17 @@
+@media (min-width: $medium-size-media) and (max-width: $max-size-media) {
```
mykhalenych, line 1: why do you duplicate this code?
delete here and use only in catalog file

src/styles/utils/_variables.scss
```diff
@@ -0,0 +1,8 @@
+$max-size-media: 1024px;
+$medium-size-media: 768px;
+$min-size-media: 488px;
+
+$card-width: 200px;
+
+$border-color: #ccc;
+$header-item-color: #060b35;
```
mykhalenych: Screenshot 2023-03-07 at 17 29 44
check your configs rules, you missed "add new line at the end"

mykhalenych (changes requested):

src/styles/utils/_media.scss
VitaliyBondarenko1982: Please, fix comments about remove duplicate code. Remove this file, you already use it in catalog

VitaliyBondarenko1982 (changes requested):

src/styles/blocks/catalog.scss
```diff
@@ -0,0 +1,37 @@
...
+  @media (min-width: $min-size-media) and (max-width: $medium-size-media) {
+    .catalog {
+      grid-template-columns: repeat(2, $card-width);
+    }
+  }
+  
+  @media (max-width: $min-size-media) {
+    .catalog {
+      grid-template-columns: repeat(1, $card-width);
+    }
```
Anastasiia-Svintsova, line 26: You can simplify it. Check if it'll work
```suggestion
   @media (min-width: $min-size-media) {
    .catalog {
      grid-template-columns: repeat(2, $card-width);
    }
  }
  
  @media (min-width: $medium-size-media) {
    .catalog {
      grid-template-columns: repeat(3, $card-width);
    }
  }
  
  @media (min-width:  $max-size-media) {
    .catalog {
      grid-template-columns: repeat(4, $card-width);
    }
```

Anastasiia-Svintsova (changes requested):
- The previous comments are not fixed. Maybe you forgot to commit and push changes? If you have any questions, please, ask them in the fe_chat 🙏 

- Align the cards container to the center 
image

- There should be four columns starting at 1024px
image

- There should be three columns starting at 768px
image

- Same with other queries. Check it, please

PR #1634, opened 2023-03-07
Develop

KateLis84:
- [DEMO LINK](https://KateLis84.github.io/layout_catalog_grid/)


loralevitska (changes requested):
Fix your media queries 
image
image
this applies to all conditions, not only the ones I indicated

in general, in code, it is better to start writing all styles from the mobile version but in this task it is up to you

PR #1639, opened 2023-03-08
catalog grid

edgarlokys:
https://edgarlokys.github.io/layout_catalog_grid/

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -17,35 +17,35 @@
   </head>
   <body>
     <header class="header">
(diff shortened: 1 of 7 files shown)


OlEzhA0 (changes requested):
hey! please check requirements 

you need to modify only this file
Screenshot 2023-03-09 at 10 48 13 AM
but you can add file for mixin

IvanFesenko (approved):

PR #1641, opened 2023-03-09
task complete

DanielSorokowski:
- [DEMO LINK](https://DanielSorokowski.github.io/layout_catalog_grid/)

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -1,13 +1,17 @@
 $card-width: 200px;
+$media-phones: 488px;
+$media-tablets: 768px;
+$media-desktop: 1024px;
 
 .catalog {
   display: grid;
+  grid-template-columns: repeat(var(--cards-number), $card-width);
   justify-items: center;
   justify-content: center;
   gap: 48px;
 
   &__card {
-    display: grid;
+    display: flex;
     justify-content: center;
     align-items: center;
     width: $card-width;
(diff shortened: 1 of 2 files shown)


src/styles/_catalog.scss
```diff
@@ -1,22 +1,36 @@
...
-  margin: 0 auto;
+.catalog {
+  display: grid;
+  justify-items: center;
+  justify-content: center;
+  gap: 48px;
 
   &__card {
-    display: flex;
+    display: grid;
```
solaryasha: Do we need to set a grid for the card ?

src/styles/_catalog.scss
```diff
@@ -1,22 +1,36 @@
...
     border-radius: 5px;
-    margin-bottom: 48px;
   }
+}
+
+@media (min-width: 488px) {
+  .catalog {
+    grid-template-columns: repeat(2, $card-width);
+  }
+}
```
solaryasha: You could wrap the media-queries in the SASS
Example:
```
.catalog {
  /* some rules */
  @media(min-width: 488px) {
    /* media-query rule */
  }
}

File: src/styles/_catalog.scss [hunk 3]
```diff
@@ -1,22 +1,36 @@
...
+  .catalog {
+    grid-template-columns: repeat(3, $card-width);
+  }
+}
 
-  &__card:not(:nth-child(4n)) {
-    margin-right: 48px;
+@media (min-width: 1024px) {
+  .catalog {
+    grid-template-columns: repeat(4, $card-width);
```
solaryasha: Also, one more trick - you could set the CSS variable for the number of cards and change in the media query.
Example:
```
--cards-number: 4;

.catalog {
  grid-template-columns: repeat(var(--cards-number), $card-width);
  
  @media (min-width: 488px) {
     --cards-number: 2;
  }
   /* and so on */
}
```
More info [here](https://www.w3schools.com/css/css3_variables_mediaqueries.asp)

src/styles/_catalog.scss
```diff
@@ -1,22 +1,36 @@
...
   }
+}
+
+@media (min-width: 488px) {
+  .catalog {
+    grid-template-columns: repeat(2, $card-width);
+  }
+}
+
+@media (min-width: 768px) {
```
solaryasha: Also,  a common practice to move media-queries breakpoints to the variables. heh

solaryasha (changes requested):

src/styles/_catalog.scss
DanielSorokowski: I know we dont, flex is fine but in the task is said "Replace flex with grid"

solaryasha (approved):
Well done!

PR #1652, opened 2023-03-12
layout_catalog_grid

YBM3:
- [DEMO LINK](https://ybm3.github.io/layout_catalog_grid/)


danielpopek94 (changes requested):
Media Queries are not working properly. Thresholds are shifted by 1px.

> There should be four columns **starting at** 1024px
> There should be three columns **starting at** 768px
> There should be two columns **starting at** 488px
> There should be one column **below** 488px

danielpopek94 (approved):

d-wasilewski (approved):

PR #1657, opened 2023-03-13
catalog grid done

ghostargg:
- [DEMO LINK](https://ghostargg.github.io/layout_catalog_grid/)

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -1,14 +1,10 @@
-@import "./utils/functions";
 @import "./utils/variables";
 
 .catalog {
   display: grid;
   grid-template-columns: repeat(4, $card-width);
   gap: 48px;
-  justify-items: center;
-
-  max-width: 1024px;
-  margin: 0 auto;
+  justify-content: center;
 
   &__card {
     display: flex;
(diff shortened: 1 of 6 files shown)


DanielSorokowski (changes requested):
- Gap between cards should be `48px`  
check gaps when cards are wraped

DanielSorokowski (approved):

src/styles/_catalog.scss
```diff
@@ -1,22 +1,30 @@
...
-    margin-right: 48px;
+  @media (max-width: 1024px) {
+    grid-template-columns: repeat(3, $card-width);
+  }
+
+  @media (max-width: 768px) {
+    grid-template-columns: repeat(2, $card-width);
+  }
+
+  @media (max-width: 488px) {
```
d-wasilewski: image
There should be 2 already :)

src/styles/_catalog.scss
```diff
@@ -1,22 +1,30 @@
...
 
-  &__card:not(:nth-child(4n)) {
-    margin-right: 48px;
+  @media (max-width: 1024px) {
+    grid-template-columns: repeat(3, $card-width);
+  }
+
+  @media (max-width: 768px) {
+    grid-template-columns: repeat(2, $card-width);
+  }
```
d-wasilewski: Also, one more trick - you could set the SCSS variable for the number of cards and change in the media query.
Example:
```
--cards-number: 4;

.catalog {
  grid-template-columns: repeat(var(--cards-number), $card-width);
  
  @media (min-width: 488px) {
     --cards-number: 2;
  }
   /* and so on */
}
```
More info [here](https://www.w3schools.com/css/css3_variables_mediaqueries.asp)

d-wasilewski (changes requested):

src/styles/_catalog.scss
ghostargg: This is really cool, code looks a lot cleaner. 

>  you could set the SCSS variable for the number of cards  

Pretty sure its typo since your example using CSS variable, but i have to write it anyway for someone who may check this code for reference. SCSS variable can only have one value because its being converted to CSS.

michalcywka (approved):

PR #1662, opened 2023-03-14
catalog grid task is solved

AllaSerhiienko:
- [DEMO LINK](https://AllaSerhiienko.github.io/layout_catalog_grid/)

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -1,22 +1,28 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
+  display: grid;
+  justify-content: center;
+  grid-template-columns: repeat(4, 200px);
+  gap: 48px;
 
-  max-width: 944px;
-  margin: 0 auto;
+  @media (max-width: 1023px) {
+    grid-template-columns: repeat(3, 200px);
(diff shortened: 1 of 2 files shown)


src/styles/_catalog.scss
```diff
@@ -1,22 +1,28 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
+  display: grid;
+  justify-content: center;
+  grid-template-columns: repeat(4, 200px);
```
ivbutsykin, line 4: I can suggest create variable for `200px` (for ex. `$card-width` and reuse it everywhere.

src/styles/_catalog.scss
```diff
@@ -1,22 +1,28 @@
...
+    grid-template-columns: 200px;
+  }
 
   &__card {
-    display: flex;
+    display: grid;
     justify-content: center;
     align-items: center;
     width: 200px;
     height: 300px;
```
ivbutsykin, line 24: Since you will have `$card-width` variable you can also create `$card-height` variable and use it here.

ivbutsykin (approved):
Good job! 
Don't forget use variables)

PR #1672, opened 2023-03-16
add task solution

Vlad-Prytula:
- [DEMO LINK](https://vlad-prytula.github.io/layout_catalog_grid/)


OleziO (changes requested):
Make header styles as in the design (add highlight and hover effect)
image

OleziO (changes requested):
Make header styles as in the design (add highlight and hover effect)
image

package.json
```diff
@@ -21,7 +21,7 @@
     "@mate-academy/bemlint": "^0.1.1",
     "@mate-academy/eslint-config": "*",
     "@mate-academy/linthtml-config": "0.0.2",
-    "@mate-academy/scripts": "^0.9.2",
+    "@mate-academy/scripts": "^1.2.8",
```
SerhiiKirik, line 24: In the future - Do not add extra files to PR
To add the file to a commit use git add src/.

SerhiiKirik (approved):
LGTM 🔥

PR #1673, opened 2023-03-17
Solution

superzuc:
https://superzuc.github.io/layout_catalog_grid/


src/styles/_catalog.scss
```diff
@@ -1,22 +1,31 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
-
-  max-width: 944px;
-  margin: 0 auto;
+  display: grid;
+  grid-template-columns: repeat(4, 200px);
+  grid-template-rows: repeat(2, 300px);
```
agniya-i, line 4: you should not specify rows, better to set height for card

src/styles/_catalog.scss
```diff
@@ -1,22 +1,31 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
-
-  max-width: 944px;
-  margin: 0 auto;
+  display: grid;
+  grid-template-columns: repeat(4, 200px);
```
agniya-i, line 3: it would be nice to add a variable for card width value

agniya-i (changes requested):

witflash (approved):
Well done 👍

PR #1675, opened 2023-03-18
done

Timur-Rudenko:
https://Timur-Rudenko.github.io/layout_catalog_grid/

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -17,35 +17,35 @@
   </head>
   <body>
     <header class="header">
-      <a class="header__logo" href="1">
+      <a class="header__logo" href="#">
         <img src="./images/logo.png" alt="Moyo logo">
       </a>
 
       <nav class="nav">
         <ul class="nav__list">
           <li class="nav__item nav__item--is-active">
-            <a href="2" class="nav__link">Apple</a>
+            <a href="apple" class="nav__link">Apple</a>
           </li>
           <li class="nav__item">
-            <a href="3" class="nav__link">Samsung</a>
+            <a href="samsung" class="nav__link">Samsung</a>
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -17,35 +17,35 @@
   </head>
   <body>
     <header class="header">
-      <a class="header__logo" href="">
+      <a class="header__logo" href="1">
```
Anastasiia-Svintsova: It's a strange href. It won't work. Fix it everywhere 
```suggestion
      <a class="header__logo" href="#">
```

https://user-images.githubusercontent.com/99403198/226103380-3f2eaf25-e0e8-4752-b3b9-beb7851faf09.mov

Anastasiia-Svintsova (changes requested):
- Why do you need a margin if you have a gap already? 
image

- Align the cards container to the center
image

src/styles/_catalog.scss
```diff
@@ -1,6 +1,20 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
+  display: grid;
+  align-items: center;
+  gap: 48px;
+  grid-template-columns: 100%;
```
SerhiiKirik, line 5: create a variable for card width
after -
Make it with card width, without `%`
```suggestion
  grid-template-columns: $cardWidth;
```

SerhiiKirik (changes requested):
Well done 🚀 
Let's make some improvements 🧑‍💻

src/styles/_catalog.scss
SerhiiKirik: you have 
image


should be
image

PR #1681, opened 2023-03-20
catalog grid solution v1

bjeziorek:
- [DEMO LINK](https://bjeziorek.github.io/layout_catalog_grid/)


Ozwiena (approved):

michalcywka (changes requested):
After making the page smaller, there is no distance between the card and the left border of the browser

michalcywka (changes requested):
There should be three columns **starting** at 768px

michalcywka (approved):

PR #1684, opened 2023-03-20
first version

Ozwiena:
- [DEMO LINK](https://ozwiena.github.io/layout_catalog_grid/)

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -1,9 +1,26 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
+  display: grid;
+  gap: 48px;
+  padding: 20px;
 
   max-width: 944px;
   margin: 0 auto;
+
+  @media (max-width: 487px) {
+    grid-template-columns: repeat(1, 1fr);
+  }
+
+  @media (min-width: 488px) {
+    grid-template-columns: repeat(2, 1fr);
+  }
+
+  @media (min-width: 768px) {
+    grid-template-columns: repeat(3, 1fr);
+  }
+
+  @media (min-width: 1024px) {
+    grid-template-columns: repeat(4, 1fr);
+  }
 
   &__card {
     display: flex;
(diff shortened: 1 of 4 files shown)


src/styles/_catalog.scss
```diff
@@ -1,10 +1,27 @@
...
+  display: grid;
+  gap: 48px;
+  padding: 20px;
 
   max-width: 944px;
   margin: 0 auto;
 
+  @media (max-width: 487px) {
+    grid-template-columns: repeat(1, 1fr);
+  }
```
Amsko6: You can put this template columns in .catalog and delete this @media, but i think it's just cosmetics.

src/styles/_catalog.scss
```diff
@@ -15,8 +32,4 @@
     border-radius: 5px;
     margin-bottom: 48px;
```
Amsko6: With this margin bottom, empty space between each row equals 96px (gap + margin), gap only should be enough.

src/styles/_catalog.scss
```diff
@@ -1,10 +1,27 @@
...
 
   max-width: 944px;
   margin: 0 auto;
 
+  @media (max-width: 487px) {
+    grid-template-columns: repeat(1, 1fr);
+  }
+
+  @media (min-width: 488px) {
+    grid-template-columns: repeat(2, 1fr);
```
Amsko6: Gaps between cards are different depending on page size, if i understand task correctly they should not. It may be becuase of using 1fr instead of 200px in column size, but i'm not sure.

Amsko6 (changes requested):

Amsko6 (approved):
:)

src/styles/_catalog.scss
```diff
@@ -1,9 +1,22 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
-
+  display: grid;
```
michalcywka, line 2: We can make it a bit nicer by adding with resolutions < 1024 by adding:
```
justify-content: center;
```

michalcywka (changes requested):
Small cosmetic change :)

michalcywka (approved):

PR #1686, opened 2023-03-20
add task solution

marshalek37:
[DEMO LINK](https://marshalek37.github.io/layout_catalog_grid/)


DanielSorokowski (changes requested):
In this task, you need to just make a grid with given cards don't copy code from previous tasks just edit _catalog.scss from flex to grid.

PR #1691, opened 2023-03-21
Develop

6houston6:
- [DEMO LINK](https://6houston6.github.io/layout_catalog_grid/)


polosanya (changes requested):
Cards should always be centred
image
image

PR #1700, opened 2023-03-22
first task approach 22/03/2023 19:21

mwgreybeard72:
[DEMO LINK](https://mwgreybeard72.github.io/layout_catalog_grid/)

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -1,11 +1,12 @@
+@import "./utils/variables.scss";
+
 .catalog {
   display: grid;
-  grid-template-columns: 200px;
+  justify-content: center;
+  grid-template-columns: $column-size;
   gap: 48px;
-
+  
   max-width: 944px;
-
-  margin: 0 auto;
 
   &__card {
     display: flex;
@@ -13,7 +14,7 @@
     align-items: center;
 
     height: 300px;
-    width: 200px;
+    width: $column-size;
 
     border: 1px solid #ccc;
     border-radius: 5px;
@@ -22,24 +23,18 @@
 
 @media (min-width: 1024px) {
   .catalog {
-    grid-template-columns: repeat(4, 200px);
+    grid-template-columns: repeat(4, $column-size);
   }
 }
 
 @media (min-width: 768px) and (max-width: 1023px) {
   .catalog {
-    grid-template-columns: repeat(3, 200px);
+    grid-template-columns: repeat(3, $column-size);
   }
 }
 
 @media (min-width: 488px) and (max-width: 767px) {
   .catalog {
-    grid-template-columns: repeat(2, 200px);
-  }
-}
-
-@media (max-width: 487px) {
-  .catalog {
-    grid-template-columns: repeat(1, 200px);
+    grid-template-columns: repeat(2, $column-size);
   }
 }
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -5,8 +5,20 @@
   justify-content: center;
   grid-template-columns: $column-size;
   gap: 48px;
-  
+
   max-width: 944px;
+
+  @media (min-width: 488px) and (max-width: 767px) {
+      grid-template-columns: repeat(2, $column-size);
+  }
+
+  @media (min-width: 768px) and (max-width: 1023px) {
+      grid-template-columns: repeat(3, $column-size);
+  }
+
+  @media (min-width: 1024px) {
+      grid-template-columns: repeat(4, $column-size);
+  }
 
   &__card {
     display: flex;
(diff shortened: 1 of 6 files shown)


src/styles/_catalog.scss
```diff
@@ -1,22 +1,45 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
+  display: grid;
+  grid-template-columns: 200px;
+  gap: 48px;
 
   max-width: 944px;
+
   margin: 0 auto;
```
PufAnimalsDev: Shouldn't be here. Check grid properties. Hint: similar to flex.

src/styles/_catalog.scss
```diff
@@ -1,22 +1,45 @@
...
 
   &__card {
     display: flex;
     justify-content: center;
     align-items: center;
-    width: 200px;
+
     height: 300px;
+    width: 200px;
```
PufAnimalsDev: use vairable, you have a lot of 200px

src/styles/_catalog.scss
```diff
@@ -1,22 +1,45 @@
...
+  .catalog {
+    grid-template-columns: repeat(2, 200px);
+  }
+}
 
-  &__card:not(:nth-child(4n)) {
-    margin-right: 48px;
+@media (max-width: 487px) {
+  .catalog {
+    grid-template-columns: repeat(1, 200px);
```
PufAnimalsDev: this is the repetition of styles. You use the same propery a lot of times. Consider how to do it with a variable. Maybe you should consider a mobile-first approach

src/styles/_catalog.scss
```diff
@@ -1,22 +1,45 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
+  display: grid;
+  grid-template-columns: 200px;
```
PufAnimalsDev: this is redundant if you want one column

PufAnimalsDev (changes requested):
I suggest changes because I'm sure it won't pass

src/styles/container.scss
```diff
@@ -0,0 +1,6 @@
+.container {
+
+  &__catalog {
+    margin: 0 auto;
+  }
+}
```
PufAnimalsDev, line 6: it's not a bug, but take a look at this grid property: 
https://learncssgrid.com/#aligning-grid-items
I will mention that the content of the task only mentioned the change in _catalog.scss

src/styles/_catalog.scss
```diff
@@ -1,22 +1,34 @@
...
+
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
+  display: grid;
+  justify-content: center;
+  grid-template-columns: $column-size;
+  gap: 48px;
 
   max-width: 944px;
```
PufAnimalsDev, line 9: Ok, I don't think it's a big mistake but you have set the width of the card 200px and you have set width gap 48px. 
So 4x200 + 3x48 = 944 and hypothetically suppose that the client no longer wants 4 columns but 5 on large screens. Then you change max-width becouse heve now 5x200 + 4x48 and in "@media (min-width: 1024px)" you have to give 5. So in my opinion it's better to remove it. Because if someone does not know that there is a max-width here we will have it: https://prnt.sc/fD-hmzLe7lVR

src/styles/_catalog.scss
```diff
@@ -1,22 +1,34 @@
...
+      grid-template-columns: repeat(2, $column-size);
+  }
+
+  @media (min-width: 768px) and (max-width: 1023px) {
+      grid-template-columns: repeat(3, $column-size);
+  }
+
+  @media (min-width: 1024px) {
+      grid-template-columns: repeat(4, $column-size);
+  }
```
PufAnimalsDev, line 21: Better, but you can remember that in SCSS you can use CSS. So you cam watch 
[PL_FE_MAR23_LESSON 20_03_2023.mp4] from 1:29:27 or try learn more about each https://sass-lang.com/documentation/at-rules/control/each . There you can click always on "=> CSS" and you will see how it compiles to css. IT is very helpful.

src/styles/_catalog.scss
```diff
@@ -1,22 +1,34 @@
...
 
   &__card {
     display: flex;
     justify-content: center;
     align-items: center;
-    width: 200px;
+
```
PufAnimalsDev, line 27: missed it before, there should be a grid here

PufAnimalsDev (approved):
Great job of course I approve, but be sure to change card element to grid, you won't get approved without it. The rest of the comments are suggestions. Maybe it will pass, change it if you want. Your decision. Hava a nice weekend

src/styles/_catalog.scss
michalcywka: It's more about main the main container :)

src/styles/_catalog.scss
michalcywka: Good point, but for this task this solution is enough :)

src/styles/_catalog.scss
michalcywka: Also good point, but for this task it's good enough :)

michalcywka (approved):

PR #1707, opened 2023-03-23
grid catalog

Misha408:
- [DEMO LINK](https://misha408.github.io/layout_catalog_grid/)


src/styles/_media.scss
```diff
@@ -0,0 +1,20 @@
+@media (max-width: 1023px) {
+  .catalog {
+    grid-template-columns: repeat(3, 200px);
```
vlad-demchuk: would be good to move width to a variable

vlad-demchuk (changes requested):
use variables for the main values (don't overuse variables) so that you'll be able to reuse them and give them descriptive names

src/styles/_media.scss
YegorKorenkov: not implemented

src/styles/_media.scss
```diff
@@ -0,0 +1,20 @@
...
+    width: 600px;
```
YegorKorenkov: why you hardcode width? Please check it
image

src/styles/_catalog.scss
```diff
@@ -1,22 +1,17 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
-
-  max-width: 944px;
+  display: grid;
```
YegorKorenkov, line 4: Please also centering cards inside catalog

YegorKorenkov (changes requested):

YegorKorenkov (approved):
Looks good, please don't forget re-deploy after making changes

PR #1724, opened 2023-03-27
catalog grid 1.0

IzzetYusufov:
- [DEMO LINK]( https://IzzetYusufov.github.io/layout_catalog_grid/ )

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -1,22 +1,31 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
-
-  max-width: 944px;
+  display: grid;
+  justify-content: center;
+  gap: 48px;
   margin: 0 auto;
+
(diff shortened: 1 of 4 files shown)


src/styles/_catalog.scss
```diff
@@ -1,22 +1,31 @@
...
-  flex-wrap: wrap;
-
-  max-width: 944px;
+  display: grid;
+  justify-content: center;
+  gap: 48px;
   margin: 0 auto;
 
+  @media (max-width: 487px) {
+    grid-template-columns: 200px;
```
YegorKorenkov, line 8: it is better to put frequent repetitive values into a variable, this is necessary in order to, for example, change the width of the card, and not to look for all the places where you set it, but to change it only in one place (change the value of the variable)

YegorKorenkov (approved):
Looks good, but read the comment and try to remember it

PR #1731, opened 2023-03-30
added catalog grid

helenmarth:
- [DEMO LINK](https://helanmarth.github.io/layout_catalog_grid/)


VitaliyBondarenko1982 (approved):
link does not work, 
may be some delay on gh pages
but code looks like good

PR #1738, opened 2023-04-03
catalogGrid

oleksandr-rudiuk:
- [DEMO LINK](https://oleksandr-rudiuk.github.io/layout_catalog_grid/)

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -1,9 +1,19 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
+  grid-template-columns: repeat(1, 1fr);
+  display: grid;
+  gap: 48px;
 
-  max-width: 944px;
-  margin: 0 auto;
+  $wides:
+    "2" 488px,
+    "3" 768px,
+    "4" 1024px;
+
+  @each $i,
+    $wide in $wides {
+    @media (min-width:#{$wide}) {
+      grid-template-columns: repeat(#{$i}, 1fr);
+    }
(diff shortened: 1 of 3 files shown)


src/styles/_catalog.scss
```diff
@@ -1,9 +1,19 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
+  grid-template-columns: repeat(1, 1fr);
```
DarkMistyRoom: no need to use repeat for one column. 🤔

src/styles/_catalog.scss
```diff
@@ -1,9 +1,19 @@
...
+  display: grid;
+  gap: 48px;
 
-  max-width: 944px;
-  margin: 0 auto;
+  $wides:
```
DarkMistyRoom:
```suggestion
  $widths:
```

DarkMistyRoom (changes requested):
The gap between cards should be 48px and the catalog block should be centered
image
Here is an example
image

src/styles/_catalog.scss
```diff
@@ -1,22 +1,30 @@
...
+  justify-content: center;
   margin: 0 auto;
 
+  $widths:
+    "2" 488px,
+    "3" 768px,
+    "4" 1024px;
+
+  @each $i,
+    $wide in $widths {
```
loralevitska, line 14:
```suggestion
    $width in $widths {
```

loralevitska (approved):

PR #1747, opened 2023-04-06
ready

danchkk:
- [DEMO LINK](https://danchkk.github.io/layout_catalog_grid/)


src/styles/_catalog.scss
```diff
@@ -1,9 +1,11 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
-
+  display: grid;
+  grid-template-columns: repeat(4, 200px);
+  grid-template-rows: repeat(2, 300px);
```
vlad-demchuk: do we need `grid-template-rows`?

vlad-demchuk: also, you could move `width` to a variable since you use it many time

vlad-demchuk (changes requested):

etojeDenys (approved):
LGTM

PR #1758, opened 2023-04-11
Develop

zisuan-od:
https://zisuan-od.github.io/layout_catalog_grid/


OleziO (changes requested):
The link must not refresh the page. Make links hover effect and make links more meaningful.

zisuan-od:
> The link must not refresh the page. Make links hover effect and make links more meaningful.

image

In description to task is written "don't modify header"
Links are displaced in the header. I didn't change the header.

Is it wrong?

src/styles/blocks/_catalog.scss
```diff
@@ -0,0 +1,29 @@
+.catalog {
+  display: grid;
+  grid-template-columns: repeat(4, auto);
+  justify-content: center;
+  gap: 48px;
+
+  @media (max-width: 1024px) {
```
anstsot: There should be four columns starting at 1024px. so it needs to be fixed
```suggestion
  @media (max-width: 1023px) {
```

anstsot: please check other media queries

anstsot (changes requested):
It’s almost done! Let’s improve a little

src/styles/blocks/_catalog.scss
zisuan-od: image
done

nazarmatsevych (approved):

PR #1765, opened 2023-04-14
add task solution

Dolnys:
- [DEMO LINK](https://Dolnys.github.io/layout_catalog_grid/)


anpiet (changes requested):
It's just a small mistake, but: 
There should be four columns **starting at** `1024px`
  - There should be three columns **starting at** `768px`
  - There should be two columns **starting at** `488px`
  - There should be one column below `488px`

On your website it starts from +1px:
- one column is below 489px
- two columns starts at 489px
- three columns at 769px
- four columns at 1025px

Dolnys:
I reduced the media queries by 1px, now it should be fine

PR #1771, opened 2023-04-17
add solution

PashaD9:
- [DEMO LINK](https://PashaD9.github.io/layout_catalog_grid/)


src/styles/main.scss
```diff
@@ -10,5 +10,6 @@ body {
   box-sizing: border-box;
 }
 
+@import "_variables.scss";
```
tetiana-ivanchenko: It's good to see that you've imported '_variables.scss', but it should be imported before other SCSS files to use these variables effectively in other files. Please move the import line for '_variables.scss' to the top of the imports list. 😊

tetiana-ivanchenko (approved):

PR #1792, opened 2023-04-27
add grid positioning to layout_catalog task

cephy77:
[DEMO LINK](https://cephy77.github.io/layout_catalog_grid/)


src/styles/utils/_variables.scss
```diff
@@ -0,0 +1,21 @@
+$color-blue-accent: #00acdc;
+$color-main-accent: #060b35;
+$color-secondary: #616070;
+$color-white: #fff;
```
generatedman, line 4: it's redundant cause white can't be changed

generatedman (approved):
good job, almost perfect!

src/styles/utils/_variables.scss
cephy77: @generatedman Yes, sounds weird, like white is always white and so does black. But I mean that there are lots of shades of white, and In that particular project white is #fff and not #fefcf3. Maybe I should name it color-light or something.

src/styles/utils/_variables.scss
```diff
@@ -0,0 +1,21 @@
...
+$screen-small: 488px;
+$screen-medium: 768px;
+$screen-large: 1024px;
+
+$gap: 48px;
+
+$one-card-width: $width-card;
+$two-cards-width: $width-card * 2 + $gap;
+$three-cards-width: $width-card * 3 + $gap * 2;
+$four-cards-width: $width-card * 4 + $gap * 3;
```
silchyn, line 21: could be function

silchyn (approved):

src/styles/utils/_variables.scss
cephy77: agree

PR #1795, opened 2023-04-28
create a grid catalog

Solovei-Andrii:
[DEMO LINK](https://Solovei-Andrii.github.io/layout_catalog_grid/)


vlad-demchuk (approved):

vlad-demchuk (changes requested):
cards width should be 200px
image

PR #1796, opened 2023-04-28
replace flex in grid

GlibVK:
https://glibvk.github.io/layout_catalog_grid/


DarkMistyRoom (changes requested):
there should be maximum 4 columns but we got 6
image
3 columns for 1023px
image

Please check all these requirements
image

PR #1802, opened 2023-04-28
add task solution layout-grid

annholovko:
[DEMO LINK](https://annholovko.github.io/layout_catalog_grid/)


andrusb21 (approved):
Good job 🔥🔥🔥

src/utils/_mixins.scss
```diff
@@ -0,0 +1,15 @@
+@mixin grid($columns, $screen-size, $card-size,) {
+  @media (min-width: $screen-size) {
+    grid-template-columns: repeat($columns, $card-size);
+  }
+}
+
+@mixin image-size($size) {
+  width: $size;
+  height: $size;
+}
```
Ihor-Karpyn:
```suggestion
@mixin square($size) {
  width: $size;
  height: $size;
}
```

src/utils/_mixins.scss
```diff
@@ -0,0 +1,15 @@
...
+
+@mixin flex {
+  display: flex;
+  align-items: center;
+}
```
Ihor-Karpyn: If mixin has name `flex` it should include only flex
```suggestion
@mixin flex {
  display: flex;
}
```

src/utils/_mixins.scss
```diff
@@ -0,0 +1,15 @@
+@mixin grid($columns, $screen-size, $card-size,) {
+  @media (min-width: $screen-size) {
+    grid-template-columns: repeat($columns, $card-size);
+  }
+}
```
Ihor-Karpyn: Mixin `grid` should include `display: grid;` This mixin worsens code readability. My suggestion is to remove it and use media queries inside the class.

To simplify writing media add mixins like that

```
@mixin onTablet {
  @media (max-width: 768px) {
    @content
  }
}
```

Ihor-Karpyn (approved):
GJ, left few suggestions

src/utils/_mixins.scss
annholovko: all suggestions were implemented

PR #1804, opened 2023-04-30
Task solution

julshvets:
- [DEMO LINK](https://julshvets.github.io/layout_catalog_grid/)

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -1,22 +1,40 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
+  display: grid;
+  grid-template-columns: $card-width;
+  grid-auto-rows: $card-height;
+  gap: 48px;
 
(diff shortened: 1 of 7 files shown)


src/styles/_catalog.scss
```diff
@@ -1,22 +1,40 @@
...
+    & {
+      grid-template-columns: repeat(3, $card-width);
+    }
+  }
+
+  @media (min-width: 1024px) {
+    & {
+      grid-template-columns: repeat(4, $card-width);
+    }
+  }
```
okosohor: you can use `1fr` instead of $card with

okosohor (approved):
good job

src/styles/_catalog.scss
naziamoff: yeah, good point

naziamoff (approved):
looks good 👍

PR #1807, opened 2023-04-30
Develop

snuber-w:
- [DEMO LINK](https://snuber-w.github.io/layout_catalog_grid/)

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -1,9 +1,29 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
+  display: grid;
+  margin: auto;
+  justify-content: space-evenly;
+  gap: 48px;
 
-  max-width: 944px;
-  margin: 0 auto;
+  @media (min-width: 488px) {
+    & {
+      max-width: 488px;
+      grid-template-columns: repeat(2, 200px);
+    }
+  }
+
+  @media (min-width: 768px) {
+    & {
+      max-width: 768px;
+      grid-template-columns: repeat(3, 200px);
+    }
+  }
+  
+  @media (min-width: 1024px) {
+    & {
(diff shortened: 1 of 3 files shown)


src/styles/_catalog.scss
```diff
@@ -1,9 +1,29 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
+  display: grid;
+  margin: auto;
+  justify-content: space-evenly;
```
IrynaBalandiukh: I think it's better to use justify-content: center and delete margin

Check the vertical gaps, because the justify-content: space-evenly has made them larger than they need to be

src/styles/_catalog.scss
```diff
@@ -1,9 +1,29 @@
...
+      grid-template-columns: repeat(3, 200px);
+    }
+  }
+
+  @media (min-width: 1024px) {
+    & {
+      max-width: 1024px;
+      grid-template-columns: repeat(4, 200px);
+    }
+  }
```
IrynaBalandiukh: I think you can get rid of the max-width

IrynaBalandiukh (changes requested):
Good job, but let`s check the gaps

src/styles/_catalog.scss
```diff
@@ -1,9 +1,26 @@
...
+  display: grid;
+  margin: auto;
+  justify-content: center;
+  gap: 48px;
 
-  max-width: 944px;
-  margin: 0 auto;
+  @media (min-width: 488px) {
+    & {
+      grid-template-columns: repeat(2, 200px);
```
naziamoff, line 9: you can create a variable for the card-size as you're using it a couple of times in your code

naziamoff (approved):
looks good 👍

PR #1810, opened 2023-04-30
add task solution

amashutikov:
- [DEMO LINK](https://amashutikov.github.io/layout_catalog_grid/)


src/styles/blocks/_catalog.scss
```diff
@@ -0,0 +1,29 @@
+.catalog {
+  display: grid;
+  gap: 48px;
+  grid-template-columns: $card-width $card-width $card-width $card-width;
```
ghost: what if we need more columns? let's use `repeat()` function

src/styles/blocks/_catalog.scss
```diff
@@ -0,0 +1,29 @@
...
+  }
+
+  @media (max-width: 767px) {
+    grid-template-columns: $card-width $card-width;
+  }
+
+  @media (max-width: 487px) {
+    grid-template-columns: $card-width;
+  }
+}
```
ghost: it would be better to make mobile grids first and go to the highest resolution

ghost: also you can try to implement this functionality with a single loop

ghost (changes requested):

naziamoff (approved):
🔥

PR #1815, opened 2023-05-01
add task sollution

EugeneSherepa:
- [DEMO LINK](https://EugeneSherepa.github.io/layout_catalog_grid/)

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -1,9 +1,26 @@
 .catalog {
-  display: flex;
+  display: grid;
   flex-wrap: wrap;
 
   max-width: 944px;
   margin: 0 auto;
+
+  @media screen and (max-width: 488px) {
+    grid-template-columns: repeat(1, 1fr);
+  }
+  
+  @media screen and (min-width: 488px) {
+    grid-template-columns: repeat(2, 1fr);
+  }
+
+  @media screen and (min-width: 768px) {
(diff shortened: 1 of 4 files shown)


src/styles/_catalog.scss
```diff
@@ -1,10 +1,27 @@
...
+
+  @media screen and (min-width: 768px) {
+    grid-template-columns: repeat(3, 1fr);
+  }
+
+  @media screen and (min-width: 1024px) {
+    grid-template-columns: repeat(4, 1fr);
+  }
+
+
```
cephy77: I suggest a slight improvement. You need to eliminate card margins and use the full power of grid positioning. Use gap property. Also, try to center cards via max-wight which can be changed in media queries. Of course, there are more solutions, but I found that pretty simple)

cephy77 (changes requested):
Look nice on the desktop! Try to improve the mobile and tablet versions of it)
Знімок екрана 2023-05-01 о 16 13 09

src/styles/_catalog.scss
cephy77: I mean that grid-template-columns can be used once with repeat(auto-fill, min-max()) property

cephy77 (approved):
Noice!)😎

naziamoff (approved):
Good job 👍 

but the header is still broken on tablet

image

PR #1818, opened 2023-05-01
add task solution

KatyaZaremska:
- [DEMO LINK](https://KatyaZaremska.github.io/layout_catalog_grid/)

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -1,22 +1,30 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
+  display: grid;
+  gap: 48px;
 
   max-width: 944px;
   margin: 0 auto;
 
+
+  @media (min-width: 488px) {
+    grid-template-columns: repeat(2, 200px);
+  }
+
+  @media (min-width: 768px) {
+    grid-template-columns: repeat(3, 200px);
+  }
+
+  @media (min-width: 1024px) {
+    grid-template-columns: repeat(4, 1fr);
+  }
+
   &__card {
-    display: flex;
+    display: grid;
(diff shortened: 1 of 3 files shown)


src/styles/_catalog.scss
```diff
@@ -1,22 +1,30 @@
...
+  @media (min-width: 488px) {
+    grid-template-columns: repeat(2, 200px);
+  }
+
+  @media (min-width: 768px) {
+    grid-template-columns: repeat(3, 200px);
+  }
+
+  @media (min-width: 1024px) {
+    grid-template-columns: repeat(4, 1fr);
```
growingananas: It would be better to use variables where possible to avoid the repetition of the same values.

src/styles/_catalog.scss
```diff
@@ -1,22 +1,30 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
+  display: grid;
+  gap: 48px;
 
   max-width: 944px;
```
growingananas: Do you need max-width here?

src/styles/_catalog.scss
```diff
@@ -1,22 +1,30 @@
...
+
+  @media (min-width: 488px) {
+    grid-template-columns: repeat(2, 200px);
+  }
+
+  @media (min-width: 768px) {
+    grid-template-columns: repeat(3, 200px);
+  }
+
+  @media (min-width: 1024px) {
```
growingananas, line 19: You might use here card's width instead.

growingananas (changes requested):
Well-done! 
It would be nice to change just a few things.

silchyn (approved):

PR #1820, opened 2023-05-01
add solution

growingananas:
- [DEMO LINK](https://growingananas.github.io/layout_catalog_grid/)

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -1,22 +1,24 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
+  @extend %display-setting;
+  gap: 48px;
 
-  max-width: 944px;
-  margin: 0 auto;
+  @media (min-width: 488px) {
+    grid-template-columns: repeat(2, $card-width);
+  }
+
+  @media (min-width: 768px) {
+    grid-template-columns: repeat(3, $card-width);
+  }
+
+  @media (min-width: 1024px) {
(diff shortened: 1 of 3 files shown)


YeMyr (approved):

vitvadym (approved):
👍

src/styles/_catalog.scss
```diff
@@ -1,22 +1,24 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
+  @extend %display-setting;
```
naziamoff, line 2: What is display-setting? I have to understand what is it about by its name, but now I don't

src/styles/variables.scss
```diff
@@ -0,0 +1,8 @@
+$card-width: 200px;
+
+%display-setting {
+  display: grid;
+  justify-content: center;
+  align-items: center;
+}
```
naziamoff, line 7: should this be inside `variables` file?

src/styles/variables.scss
naziamoff, line 1: In SCSS, files that begin with an underscore (e.g., _variables.scss) are typically referred to as partials. 

So it seems to me that this file should be named with the underscore

naziamoff (approved):
looks good 👍 
see suggestions

src/styles/variables.scss
growingananas: I am not sure now...this is a placeholder. Should I keep it in catalog.scss?

PR #1821, opened 2023-05-01
add task solution

ulositska:
- [DEMO LINK](https://ulositska.github.io/layout_catalog_grid/)

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -1,22 +1,23 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
+  display: grid;
+  gap: $catalog-gap;
+  justify-content: center;
+  grid-template-columns: repeat(1, $card-width);
 
-  max-width: 944px;
-  margin: 0 auto;
+  @each $width, $col-number in $breakpoints {
+    @media (min-width: $width) {
+      grid-template-columns: repeat($col-number, $card-width);
+    }
+  }
 
   &__card {
     display: flex;
     justify-content: center;
     align-items: center;
-    width: 200px;
+    width: $card-width;
     height: 300px;
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
-  }
-
-  &__card:not(:nth-child(4n)) {
-    margin-right: 48px;
   }
 }
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -14,10 +14,9 @@
     display: flex;
     justify-content: center;
     align-items: center;
-    width: $card-width;
-    height: 300px;
+    width: 100%;
+    height: $card-height;
     border: 1px solid #ccc;
     border-radius: 5px;
-    margin-bottom: 48px;
   }
 }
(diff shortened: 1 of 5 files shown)


src/styles/_catalog.scss
```diff
@@ -1,22 +1,23 @@
...
+      grid-template-columns: repeat($col-number, $card-width);
+    }
+  }
 
   &__card {
     display: flex;
     justify-content: center;
     align-items: center;
-    width: 200px;
+    width: $card-width;
```
panch1811: `width: 100%` would be better here, coz your card is always inside your grid cell.

src/styles/_catalog.scss
```diff
@@ -1,22 +1,23 @@
...
     height: 300px;
     border: 1px solid #ccc;
     border-radius: 5px;
     margin-bottom: 48px;
   }
```
panch1811: do u need it?

src/styles/utils.scss/_variables.scss
```diff
@@ -0,0 +1,7 @@
+$card-width: 200px;
+$catalog-gap: 48px;
+
```
panch1811, line 4: do u need it as variable? in your code u use it only once.

src/styles/utils.scss/_variables.scss
```diff
@@ -0,0 +1,7 @@
...
+$breakpoints:
+  488px 2,
+  768px 3,
+  1024px 4;
```
panch1811, line 8: good solution! :)

panch1811 (changes requested):
image
is this gap 48px? I guess u have to fix it :)

P.S check other comments below ;)

src/styles/_catalog.scss
ulositska: Oww thanks that's very very good advice:fire:

src/styles/_catalog.scss
ulositska: Initially, I wanted to have a bigger distance between rows, but you're totally right. Cause, even in that case, I could have used gap with 2 values:facepalm: Fixed!

src/styles/utils.scss/_variables.scss
ulositska: Valid! But I want to keep it as a variable for now. I'm not sure if it's considered bad practice, but I enjoy having them all in one place, hypothetically the value may change in the future, so it's easier for me this way.

src/styles/main.scss
```diff
@@ -10,5 +10,6 @@ body {
   box-sizing: border-box;
 }
 
+@import "./utils.scss/variables";
```
naziamoff, line 13: the file is named with underscore, but you importing it without 👀

src/styles/_catalog.scss
```diff
@@ -1,22 +1,22 @@
...
+  }
 
   &__card {
     display: flex;
     justify-content: center;
     align-items: center;
-    width: 200px;
-    height: 300px;
+    width: 100%;
+    height: $card-height;
```
naziamoff, line 18: usually we don't set exact height for such elements, we let the content dictate it. I don't see the use case except for when we're mocking cards like in this task. Not a mistake though, but consider it

naziamoff (approved):
👍

PR #1822, opened 2023-05-02
Completed the catalog grid task

Shevtsooov:
- [DEMO LINK](https://Shevtsooov.github.io/layout_catalog_grid/)

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -1,9 +1,25 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
+  display: grid;
+  justify-content: center;
+
+  grid-template-columns: repeat(auto-fill, 200px);
+  gap: 48px;
 
   max-width: 944px;
   margin: 0 auto;
+  padding-bottom: 48px;
+
+  @media (max-width: 1023px) {
+    grid-template-columns: repeat(3, 200px);
+  }
+
+  @media (max-width: 767px) {
+    grid-template-columns: repeat(2, 200px);
+  }
+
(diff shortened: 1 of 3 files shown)


src/styles/_catalog.scss
```diff
@@ -1,9 +1,25 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
+  display: grid;
+  justify-content: center;
+
+  grid-template-columns: repeat(auto-fill, 200px);
+  gap: 48px;
 
   max-width: 944px;
```
SofiaLiubchenko, line 8: you can delete max-width property

src/styles/_catalog.scss
```diff
@@ -1,9 +1,25 @@
...
   margin: 0 auto;
+  padding-bottom: 48px;
```
SofiaLiubchenko, line 10: you can delete padding-bottom - there is a 'gap' instead

SofiaLiubchenko (approved):
good job!

src/styles/_catalog.scss
```diff
@@ -1,9 +1,25 @@
...
+
+  @media (max-width: 1023px) {
+    grid-template-columns: repeat(3, 200px);
```
naziamoff, line 13: would be great to create a variable for that

src/styles/_catalog.scss
naziamoff: I would say it rather looks like a margin, as it sets empty space for the whole catalog, not for some exact card where gap works just fine

naziamoff (approved):

PR #1823, opened 2023-05-02
Develop

hekaq:
- [DEMO LINK](https://hekaq.github.io/layout_catalog_grid/)


src/styles/blocks/_catalog.scss
```diff
@@ -0,0 +1,35 @@
...
+    height: 300px;
+    border: 1px solid #ccc;
+    border-radius: 5px;
+
+  }
+
+  &__card:not(:nth-child(4n)) {
+    margin-right: 48px;
+  }
+}
```
ulositska, line 35: You've already set the gap property for catalog, I think, it makes the margins for the cards redundant:broom:

src/styles/blocks/_catalog.scss
```diff
@@ -0,0 +1,35 @@
...
+    grid-template-columns: repeat(3, $card-width);
+  }
+
+  @media (max-width: 768px) {
+    grid-template-columns: repeat(2, $card-width);
+  }
+
+  @media (max-width: 488px) {
+    grid-template-columns: repeat(1, $card-width);
+  }
```
ulositska, line 19: It's an option, I'm just curious, is there a particular reason why you decided to take desktop first approach here?:thinking:

ulositska (approved):
The layout changes as supposed to! You did a good job using grid 🔥

src/styles/blocks/_catalog.scss
hekaq: there is no specific reason, I think it would be better to use mobile first approach in the future

naziamoff (approved):
looks good 👍 

One thing:

In SCSS, files that begin with an underscore (e.g., _variables.scss) are typically referred to as partials.

So it seems to me that some files should be named with the underscore

PR #1825, opened 2023-05-02
add task solution

PanyaPrimal:
[DEMO LINK](https://PanyaPrimal.github.io/layout_catalog_grid/)

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -1,22 +1,25 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
-
-  max-width: 944px;
-  margin: 0 auto;
+  display: grid;
+  gap: 48px;
+  justify-content: center;
 
   &__card {
     display: flex;
     justify-content: center;
     align-items: center;
-    width: 200px;
+    width: $card-width;
     height: 300px;
     border: 1px solid #ccc;
     border-radius: 5px;
-    margin-bottom: 48px;
   }
 
-  &__card:not(:nth-child(4n)) {
-    margin-right: 48px;
+  @media (min-width: 488px) {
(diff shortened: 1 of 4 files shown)


vakopov (approved):
Well done!

src/styles/_catalog.scss
```diff
@@ -1,22 +1,25 @@
...
-    margin-bottom: 48px;
   }
 
-  &__card:not(:nth-child(4n)) {
-    margin-right: 48px;
+  @media (min-width: 488px) {
+    grid-template-columns: repeat(2, $card-width);
+  }
+  @media (min-width: 768px) {
+    grid-template-columns: repeat(3, 200px);
```
naziamoff, line 20: Replace the fixed width value (200px) with the variable $card-width for consistency

src/styles/_catalog.scss
```diff
@@ -1,22 +1,25 @@
...
+  }
+  @media (min-width: 1024px) {
+    grid-template-columns: repeat(4, 200px);
```
naziamoff, line 23:
```suggestion
  @media (min-width: 488px) {
    grid-template-columns: repeat(2, $card-width);
  }
  
  @media (min-width: 768px) {
    grid-template-columns: repeat(3, 200px);
  }
  
  @media (min-width: 1024px) {
    grid-template-columns: repeat(4, 200px);
```

src/styles/utils/vars.scss
naziamoff, line 1: In SCSS, files that begin with an underscore (e.g., _variables.scss) are typically referred to as partials.

So it seems to me that this file should be named with the underscore, as well as in its import

naziamoff (approved):
great job 👍 
There are some suggestions though

PR #1832, opened 2023-05-02
(ym) catalog grid

YeMyr:
- [DEMO LINK](https://yemyr.github.io/layout_catalog_grid/)


src/styles/_catalog.scss
```diff
@@ -1,22 +1,29 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
+  display: grid;
+  grid-template-columns: auto;
```
ghost: i'm not sure if the behavior of an element with the value `auto` can be understood by other developers, let's set the fixed `grid-template-columns` for our grid
```suggestion
  grid-template-columns: $card-width;
```

src/styles/_catalog.scss
```diff
@@ -1,22 +1,29 @@
...
+  justify-content: center;
+  gap: 48px;
 
-  max-width: 944px;
-  margin: 0 auto;
+  @media (min-width: 488px) {
+    grid-template-columns: repeat(2, 200px);
```
ghost: let's create a variable for our grid item sizes and use it in all media queries

ghost (changes requested):

src/styles/_catalog.scss
YeMyr: yes, you are right, it will be better to add variable
thanks
fixed
Have a nice weekend)

src/styles/_catalog.scss
YeMyr: fixed

ghost (approved):
🔥

silchyn (approved):
`_catalog.scss` -> `catalog.scss`

YeMyr:
thanks, I've changed file names

PR #1839, opened 2023-05-03
Grid solution added

olehmahera:
- [DEMO LINK](https://olehmahera.github.io/layout_catalog_grid/)


src/styles/blocks/_catalog.scss
```diff
@@ -0,0 +1,26 @@
...
+  @media (min-width: 1024px) {
+    grid-template-columns: repeat(4, $card-width);
+  }
+
+  &__card {
+    display: grid;
+    place-items: center;
+    height: 300px;
+    border: 1px solid #ccc;
+    border-radius: 5px;
```
Ihor-Karpyn, line 24: Why the catalog contains styles associated with cards?

Ihor-Karpyn (changes requested):

src/styles/blocks/_catalog.scss
olehmahera: Cards are elements of the Catalog block. Do I have to create a Card block?

PR #1846, opened 2023-05-03
grid catalog

thienlaoo:
- [DEMO LINK](https://thienlaoo.github.io/layout_catalog_grid/)


src/styles/blocks/_catalog.scss
```diff
@@ -0,0 +1,53 @@
...
+  .catalog {
+    display: grid;
+    justify-content: center;
+    grid-template-columns: repeat(1, 200px);
+    gap: 48px;
+  }
+}
+
+@media (min-width:489px) and (max-width:768px) {
+  .catalog {
```
Ihor-Karpyn: It is not recommended to include a selector within a media query in the SCSS.

````scss
.catalog {
  width: 100px;
  
  @media (...) {
    width: 200px;
  }
}
````

File: src/styles/blocks/_catalog.scss [hunk 2]
```diff
@@ -0,0 +1,53 @@
...
+}
+
+@media (min-width:1024px) {
+  .catalog {
+    display: grid;
+    justify-content: center;
+    grid-template-columns: repeat(4, 200px);
+    gap: 48px;
+  }
+}
```
Ihor-Karpyn: When making changes to the rules in the media, it is important to only include the ones that require modification. Avoid duplicating styles and ensure that if you are only altering the 'grid-template-columns' rule, it is the sole one present in the media.

src/styles/blocks/_catalog.scss
```diff
@@ -0,0 +1,53 @@
...
+  &__card {
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    width: 200px;
+    height: 300px;
+    border: 1px solid #ccc;
+    border-radius: 5px;
+    margin-bottom: 48px;
+  }
```
Ihor-Karpyn: I hope that the card can be a BEM block.

It would be beneficial to move all of the styles, except for the margins that are necessary for positioning the card, to the .card BEM block.

Ihor-Karpyn (changes requested):
Almost done

src/styles/blocks/_catalog.scss
thienlaoo: I have 'catalog_card' as the element of 'catalog' BEM block, so I think 'card' cant be separate BEM block, but I can make mixin for 'card' positioning

src/styles/blocks/_catalog.scss
```diff
@@ -0,0 +1,36 @@
...
+        margin-right: 48px;
+      }
+    }
+  }
+
+  @include gridLayout;
+
+  @media (max-width:488px) {
+    grid-template-columns: repeat(1, 200px);
+  }
```
Anna-Tkachenko, line 23: These property can be set by default without media query

Anna-Tkachenko (approved):
Good job!

PR #1861, opened 2023-05-08
create grid layout

valerkameladze:
- [DEMO LINK](https://valerkameladze.github.io/layout_catalog_grid/)

### DIFF
--- a/src/styles/_catalog.scss
+++ b/src/styles/_catalog.scss
@@ -1,5 +1,7 @@
 .catalog {
-  display: flex;
+  display: grid;
+  grid-template-columns: 1fr 1fr 1fr 1fr;
+  gap: 48px;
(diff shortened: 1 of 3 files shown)


tetiana-ivanchenko (changes requested):
image
in your case it is not 48

src/styles/_catalog.scss
```diff
@@ -1,5 +1,6 @@
 .catalog {
-  display: flex;
+  display: grid;
+  grid-template-columns: 1fr 1fr 1fr 1fr;
```
TarasHoliuk, line 3: I'd use [repeat](https://developer.mozilla.org/en-US/docs/Web/CSS/repeat) here (just FYI)

TarasHoliuk (approved):

PR #1879, opened 2023-05-15
add task solution

RomanPerepelkin:
[DEMO LINK](https://RomanPerepelkin.github.io/layout_catalog_grid/)


src/styles/utils/_variables.scss
```diff
@@ -0,0 +1 @@
+$base-card-width: 200px;
```
volodymyr-soltys97, line 1: The variable declaration is correct, but it would be better to include more variables for other aspects of the layout, like gap size and media query breakpoints.

src/styles/blocks/_catalog.scss
```diff
@@ -0,0 +1,37 @@
...
+
+  &__card {
+    display: grid;
+    grid-template-columns: repeat(1, $base-card-width);
+    text-align: center;
+    align-content: center;
+    height: 300px;
+    border: 1px solid #ccc;
+    border-radius: 5px;
+    margin-bottom: 48px;
```
volodymyr-soltys97: Remove the margin-bottom from the catalog card as it's not needed due to grid gap.

src/styles/blocks/_page.scss
```diff
@@ -0,0 +1,11 @@
+html {
+  font-family: Roboto, sans-serif;
+}
+
+body {
+  margin: 0;
+}
+
+* {
```
volodymyr-soltys97: The universal selector (*) might cause performance issues, consider using a more specific selector or a CSS reset instead.

src/styles/blocks/_page.scss
```diff
@@ -0,0 +1,11 @@
+html {
```
volodymyr-soltys97: Move the 'html' and 'body' styles to a more general file, like '_base.scss', since they are not specific to the catalog page layout.

src/styles/main.scss
```diff
@@ -1,14 +1,4 @@
...
-  margin: 0;
-}
-
-* {
-  box-sizing: border-box;
-}
-
-@import "_header.scss";
-@import "_catalog.scss";
+@import "./utils/variables";
```
volodymyr-soltys97: The import for 'variables' should be changed to '_variables.scss' for consistency with the naming convention used in other imported files.

src/styles/main.scss
```diff
@@ -1,14 +1,4 @@
...
+@import "./blocks/page";
+@import "./blocks/header";
+@import "./blocks/catalog";
```
volodymyr-soltys97: The import for 'catalog' should be changed to '_catalog.scss' for consistency with the naming convention used in other imported files.

volodymyr-soltys97 (changes requested):
Good job 👍 
But let's make your code more sophisticated

volodymyr-soltys97 (approved):
Awesome 🚀

PR #1896, opened 2023-05-18
layout_catalog_grid

aamelnyk:
- [DEMO LINK](https://aamelnyk.github.io/layout_catalog_grid/)


polosanya (changes requested):
Almost done!
It would be great to center the cards horizontally
image

PR #1908, opened 2023-06-01
add task solution

spojrzenie:


src/styles/_catalog.scss
```diff
@@ -1,22 +1,30 @@
 .catalog {
-  display: flex;
-  flex-wrap: wrap;
+  padding: 10px 20px;
+  display: grid;
+  justify-content: center;
+  gap: 48px;
```
darokrk, line 6: "There should be one column below 488px" think you missed this requirement :)

just add here `grid-template-columns: repeat(1, 200px);`

darokrk (changes requested):
Small fix is required for approval :)

src/styles/_catalog.scss
spojrzenie: just thought this limit was not necessary as there was 1 column below 488px without this limit :)

src/styles/_catalog.scss
```diff
@@ -1,22 +1,31 @@
...
 
-  max-width: 944px;
-  margin: 0 auto;
+  grid-template-columns: repeat(1, 200px);
```
darokrk, line 7: You can even try to use here  'grid-template-columns' to `repeat(auto-fit, minmax(200px, 1fr));` to make the layout responsive and just leave only this breakpoint to make sure then you get max 4 columns.

`@media all and (min-width: 1024px) {
  }`

Then you can remove the rest breakpoints for the smaller screens.

`min-width: 488px` and `min-width: 768px`

darokrk (approved):
Nice 🥇  Give approval, but look at my last mention, test it by yourself and try to implement it :) it will improve code even more

